PR #1, opened 2021-12-02
Initialize operator-sdk project, api, and provide reconcile scaffolding.

rohantmp:


rohantmp:
@nbalacha , working on more scaffolding and tasks

rohantmp:
I'll create empty functions for reconciling the csidrivers, node daemonset (daemon that discovers devices, matches against filters, creates lvmd config), lvmd, etc
I'll then create subtasks for each of these that can be picked up.

leseb (commented):
Can we please not vendor unless we really need to? Thanks!

leelavg:
### - reg deployment model of CSI controller and nodeplugin

---

> node daemonset (daemon that discovers devices, matches against filters, creates lvmd config)

- will we be following similar design of deploying lvmd and nodeplugin as deployment
- there's no way out of it, so just confirming if you thought of any other design

---

- there's a possibility to deploy lvmd and nodeplugin as daemonset if we can maintain a single lvmd configmap, however it may go over 1MB limit
- that requires a change to topolvm code, mainly in the lvmd to wait for the availability of VG

---

my query is, as we aren't accepting the topolvm version/image from the CR how best we can deploy underlying CSI controller and nodeplugin with sidecars and possibility of getting away with using a single manager for multiple controllers?

nbalacha:
> ### - reg deployment model of CSI controller and nodeplugin
> > node daemonset (daemon that discovers devices, matches against filters, creates lvmd config)
> 
> * will we be following similar design of deploying lvmd and nodeplugin as deployment
> * there's no way out of it, so just confirming if you thought of any other design
> 
> * there's a possibility to deploy lvmd and nodeplugin as daemonset if we can maintain a single lvmd configmap, however it may go over 1MB limit
> * that requires a change to topolvm code, mainly in the lvmd to wait for the availability of VG
> 
> my query is, as we aren't accepting the topolvm version/image from the CR how best we can deploy underlying CSI controller and nodeplugin with sidecars and possibility of getting away with using a single manager for multiple controllers?

We can use env variables to specify the images for now and the topolvm controller can also watch the operator config map.

rohantmp:
>     will we be following similar design of deploying lvmd and nodeplugin as deployment
>     there's no way out of it, so just confirming if you thought of any other design

We could avoid having the lvmd config as a configmap and simply have it live on the node. Each node's daemon could generate it based on existing config and logic.

rohantmp:
Have updated the scaffolding

nbalacha (approved):

PR #8, opened 2021-12-06
move logger to Reconciler struct

rohantmp:
Passing the log to all nested called functions is cumbersome.
Since we only run one Reconcile() worker at a time, we can use a
struct level var.

Signed-off-by: Rohan CJ <rohantmp@gmail.com>

### DIFF
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -55,7 +55,7 @@ type LVMClusterReconciler struct {
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
 func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	r.Log = log.FromContext(ctx).WithName(ControllerName)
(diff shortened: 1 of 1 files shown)


controllers/lvmcluster_controller.go
```diff
@@ -53,19 +54,19 @@ type LVMClusterReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
 func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.FromContext(ctx).WithName(ControllerName)
-	logger.Info("reconciling", "topolvmcluster", req)
-	result, err := r.reconcile(ctx, req, logger)
+	r.Log = log.FromContext(ctx).WithName(ControllerName)
+	r.Log.Info("reconciling", "topolvmcluster", req)
```
nbalacha, line 58: This should be "lvmcluster"

rohantmp: fixed!

nbalacha (approved):

PR #11, opened 2021-12-07
Add CSIDriver

jmolmo:
Not sure about several things...

- Should we use a "unit" for the CSIDriver? If the answer is yes, we will need to solve cross reference imports ( "controllers" needs "CSIDriver", and "CSIDriver" needs "controllers", we will have the same problem with other "pkgs".. (Not sure if i understand well the pattern to use the interface for units in the `lvmcluster_controller`)

- Maybe it would be useful a customized `context` with more things .. include the `clientset `can be useful for lot of things

### DIFF
--- a/pkg/csi/csidriver.go
+++ b/pkg/csi/csidriver.go
@@ -1,0 +1,87 @@
+package csi
+
+import (
+	"context"
+	"github.com/coreos/pkg/capnslog"
+	k8sUtils "github.com/red-hat-storage/lvm-operator/pkg/util/k8s"
+	storageV1 "k8s.io/api/storage/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	csiDriverName = "topolvm.cybozu.com"
+)
+
+var (
+	logger = capnslog.NewPackageLogger("github.com/red-hat-storage/lvm-operator", "csi")
+)
+
+// CreateCSIDriverInfo Registers CSI driver by creating a CSIDriver object
+func CreateCSIDriverInfo(ctx context.Context) error {
+	clientSet, err := k8sUtils.GetClientSet(ctx)
+	if err != nil {
+		logger.Errorf("Not possible to get a valid clientSet to add %s to CSIDrivers", csiDriverName)
+		return err
+	}
+	attachRequired := false
+	podInfoOnMount := true
+	storageCapacity := true
+
+	// Create CSIDriver object
+	csiDriver := &storageV1.CSIDriver{
+		ObjectMeta: metaV1.ObjectMeta{
+			Name: csiDriverName,
+		},
+		Spec: storageV1.CSIDriverSpec{
+			AttachRequired:       &attachRequired,
+			PodInfoOnMount:       &podInfoOnMount,
+			StorageCapacity:      &storageCapacity,
+			VolumeLifecycleModes: []storageV1.VolumeLifecycleMode{storageV1.VolumeLifecyclePersistent, storageV1.VolumeLifecycleEphemeral},
+		},
+	}
+
+	// Attach CSIDriver object
+	csiDrivers := clientSet.StorageV1().CSIDrivers()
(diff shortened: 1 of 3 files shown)


sp98 (changes requested):
unit tests?

leelavg:
@jmolmo 

- a lot has to be changed, i'll try to explain the best I can
Some good to know info:
1. Every unit (vgmanager, csicontroller, csinode, csidriver) should implement the interface `resourceManager` to be cleaner for any extensions
2. If any unit doesn't want to reconcile any method (or a specific piece of) it can be left empty (ex. csidriver resource doesn't need an update)


> Should we use a "unit" for the CSIDriver?
1. yes

> "controllers" needs "CSIDriver", and "CSIDriver" needs "controllers"
1. nope, we can get away without deploying CSIDriver resource at all 😅 
2. it's presence only informs k8s that this driver supports/needs these (attachRequired, podInfoOnMount, storageCapacity) functionalities
3. in it's absence it gets the info from one of the controller plugin RPC calls

- Why all this hassle if it's not mandatory?
1. not explaining the detail, we are deploying the CSIDriver just to save a single RPC call and it's roundtrip by setting `attachRequired` to `false`

- Reg PR
1. I was told to use `controllerutil.CreateOrUpdate` for resource creation and utilise controller-runtime client interface https://github.com/kubernetes-sigs/controller-runtime/blob/3ffc700bc2a37c97acd6ce3a03c3526df8a5e489/pkg/client/interfaces.go#L100-L110
2. Owner ref will not work on this resource so main reconciler will set a finalizer on the CR and invokes CSIDriver resource deletion
3. No, we don't need to set `resourceVersion` and it may not work I guess, afaik it's for client & server confirmation that the resource isn't  mutated while we are updating it
4. We are supporting only single CR and better use the logging field available in the main reconciler

Let me know if any of the points doesn't make sense

pkg/csi/csidriver.go
```diff
@@ -0,0 +1,87 @@
...
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	csiDriverName = "topolvm.cybozu.com"
+)
+
+var (
+	logger = capnslog.NewPackageLogger("github.com/red-hat-storage/lvm-operator", "csi")
```
rohantmp, line 17: let's use the controller-runtime logger? `LVMClusterReconciler.Log`?

pkg/util/k8s/k8s.go
```diff
@@ -0,0 +1,23 @@
+package k8s
+
+import (
+	"context"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func GetClientSet(ctx context.Context) (*kubernetes.Clientset, error) {
```
rohantmp, line 10: Let's use the controller-runtime client, it reads from a cache and you supports all runtime.Objects. `LVMClusterReconciler.Client.Create/Update/Delete`

rohantmp (changes requested):
Here is an example from ocs-operator where we took the resourceManager idea from: 
The interface: https://github.com/red-hat-storage/ocs-operator/blob/bf0946e7b4d47b799d209c491f0fcfbd71a4295c/controllers/storagecluster/reconcile.go#L33

A list of implementations: https://github.com/red-hat-storage/ocs-operator/blob/bf0946e7b4d47b799d209c491f0fcfbd71a4295c/controllers/storagecluster/reconcile.go#L352

clients and loggers are available in the first argument of ensureCreated(): https://github.com/red-hat-storage/lvm-operator/blob/a72ebd4a8fb3d478ca6610216264956b3ed52a65/controllers/lvmcluster_controller.go#L126

jmolmo:
> * a lot has to be changed, i'll try to explain the best I can, some good to know info:

Perfectly explained @leelavg , Thank you for take the time to explain!!. I think we can close this PR in favor of your PR (https://github.com/red-hat-storage/lvm-operator/pull/13).

jmolmo:
@rohantmp : Thank you for the multiple references. !! :-)

jmolmo:
Closed in favor of #13

PR #13, opened 2021-12-09
csi: add csi driver and controller plugin deployment

leelavg:

### DIFF
--- a/controllers/defaults.go
+++ b/controllers/defaults.go
@@ -38,6 +38,39 @@ var (
 	CsiProvisionerImage   = GetEnvOrDefault("CSI_PROVISIONER_IMAGE")
 	CsiLivenessProbeImage = GetEnvOrDefault("CSI_LIVENESSPROBE_IMAGE")
 	CsiResizerImage       = GetEnvOrDefault("CSI_RESIZER_IMAGE")
+
+	// Topolvm deployment label
+	AppAttr = "app.kubernetes.io/name"
+
+	TopolvmControllerServiceAccount = "topolvm-controller"
+
+	// CSI Controller container and deployment names
+	CsiRegistrarContainerName            = "csi-registrar"
+	TopolvmControllerDeploymentName      = "topolvm-controller"
+	TopolvmControllerContainerName       = "topolvm-controller"
+	TopolvmCsiResizerContainerName       = "csi-resizer"
+	TopolvmCsiProvisionerContainerName   = "csi-provisioner"
+	TopolvmCsiLivenessProbeContainerName = "liveness-probe"
+
+	// CSI Controller health endpoints
+	TopolvmControllerContainerHealthzName   = "healthz"
+	TopolvmControllerContainerLivenessPort  = int32(9808)
+	TopolvmControllerContainerReadinessPort = int32(8080)
+
+	// CSI Controller resource requests/limits
+	// TODO: Reduce these values and reach optimistic values without effecting performance
+	TopolvmControllerMemRequest = "250Mi"
+	TopolvmControllerMemLimit   = "250Mi"
+	TopolvmControllerCPURequest = "250m"
+	TopolvmControllerCPULimit   = "250m"
+
+	TopolvmControllerCsiProvisionMemRequest = "100Mi"
+	TopolvmControllerCsiProvisionMemLimit   = "100Mi"
+	TopolvmControllerCsiProvisionCPURequest = "100m"
+	TopolvmControllerCsiProvisionCPULimit   = "100m"
+
+	// CSIDriver
+	TopolvmCSIDriverName = "topolvm.cybozu.com"
 )
 
 func GetEnvOrDefault(env string) string {
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -95,7 +95,10 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 // errors returned by this will be updated in the reconcileSucceeded condition of the LVMCluster
 func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMCluster) (ctrl.Result, error) {
-	resourceList := []resourceManager{}
+	resourceList := []resourceManager{
+		&csiDriver{},
+		&controllerPlugin{},
+	}
 
 	//The resource was deleted
 	if !instance.DeletionTimestamp.IsZero() {
--- a/controllers/lvmcluster_controller_test.go
+++ b/controllers/lvmcluster_controller_test.go
@@ -19,6 +19,9 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -32,9 +35,13 @@ var _ = Describe("LVMCluster controller", func() {
 
 	Context("LvmCluster reconcile", func() {
 		It("Reconciles an LvmCluster, ", func() {
-			By("Indicate setting status to ready")
+
 			ctx := context.Background()
-			lvmCluster := &lvmv1alpha1.LVMCluster{
+
+			// LVM CR
+			lvmClusterName := types.NamespacedName{Name: testLvmClusterName, Namespace: testLvmClusterNamespace}
+			lvmClusterOut := &lvmv1alpha1.LVMCluster{}
+			lvmClusterIn := &lvmv1alpha1.LVMCluster{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      testLvmClusterName,
 					Namespace: testLvmClusterNamespace,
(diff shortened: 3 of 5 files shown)


controllers/topolvm_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		// deployment is already deleted based on ownerref and just log the msg
+		if errors.IsNotFound(err) {
+			r.Log.Info("csi controller deleted", "TopolvmController", cd.Name)
+			return nil
+		}
+		r.Log.Error(err, "unable to retrieve csi controller deployment", "TopolvmController", cd.Name)
+		return err
+	}
+	return nil
```
nbalacha, line 69: The deployment should be deleted here.

controllers/defaults.go
```diff
@@ -38,6 +38,39 @@ var (
 	CsiProvisionerImage   = GetEnvOrDefault("CSI_PROVISIONER_IMAGE")
 	CsiLivenessProbeImage = GetEnvOrDefault("CSI_LIVENESSPROBE_IMAGE")
 	CsiResizerImage       = GetEnvOrDefault("CSI_RESIZER_IMAGE")
+
+	// Topolvm deployment label
+	AppAttr = "app.kubernetes.io/name"
+
+	ContollerServiceAccount = "topolvm-controller"
```
nbalacha, line 45: Typo: ControllerServiceAccount

Please rename this to TopolvmControllerServiceAccount

jmolmo (approved):
Just minor things... probably only for discussion. LGTM

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (c controllerPlugin) updateStatus(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+	// TODO: Verify the status of controller plugin deployment and set the same on CR
+	return nil
+}
+
+func getControllerDeployment(lvmCluster *lvmv1alpha1.LVMCluster) *appsv1.Deployment {
+
+	// Topolvm CSI Controller Deployment
+	var replicaOne int32 = 1
```
nbalacha, line 79: Please rename this to either replicas or numReplicas

controllers/topolvm_csi_driver.go
```diff
@@ -0,0 +1,99 @@
...
+		r.Log.Info("csi driver", "operation", result, "name", csiDriverDeployment.Name)
+	default:
+		r.Log.Error(err, "csi driver reconcile failure", "name", csiDriverDeployment.Name)
+		return err
+	}
+	return nil
+}
+
+func (c csiDriver) ensureDeleted(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+	csiDriverDeployment := &storagev1.CSIDriver{}
```
nbalacha, line 47: This is not a deployment. Please rename to csiDriver.

controllers/topolvm_csi_driver.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	// if not deleted, initiate deletion
+	if csiDriverDeployment.GetDeletionTimestamp().IsZero() {
+		if err = r.Client.Delete(ctx, csiDriverDeployment); err != nil {
+			r.Log.Error(err, "unable to delete topolvm csi driver", "TopolvmCSIDriverName", csiDriverDeployment.Name)
+			return err
+		}
+	} else if !csiDriverDeployment.GetDeletionTimestamp().IsZero() {
```
nbalacha, line 66: Just
else {
...
}

 should do here

controllers/topolvm_csi_driver.go
```diff
@@ -0,0 +1,99 @@
...
+	csiDriver := &storagev1.CSIDriver{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: TopolvmCSIDriverName,
+		},
+		Spec: storagev1.CSIDriverSpec{
+			AttachRequired:  &attachRequired,
+			PodInfoOnMount:  &podInfoOnMount,
+			StorageCapacity: &storageCapacity,
+			// topolvm supports both Ephemeral and Persistent volumes
+			VolumeLifecycleModes: []storagev1.VolumeLifecycleMode{storagev1.VolumeLifecyclePersistent, storagev1.VolumeLifecycleEphemeral},
```
nbalacha, line 95: Do we need storagev1.VolumeLifecycleEphemeral here?

controllers/topolvm_controller.go
rohantmp: For a deployment, an ownerReference should be enough for cleanup.
Not for cluster scoped objects or complex cleanups though.

nbalacha: What is the expected behaviour if the lvmcluster object is deleted? Shouldn't the deployment be deleted as well?

leelavg: - I understood that if lvm cr is deleted then controller will also be deleted due to owner ref
- However I'm unsure whether dependents will be deleted after realizing deletestamp on owner or owner needs to be deleted (like actual deletion ~Not found)🤔

nbalacha: At what point will the deployment be deleted? As soon as the delete operation is performed or after the finalizer is removed from the LvmController? The operator waits for all resources to be deleted before removing the finalizer.

If it is deleted before the finalizer is removed, then this method should return an error as long as  the deployment resource exists in order to prevent the removal of the finalizer. If not, this is not a good candidate for the resource manager.

controllers/defaults.go
```diff
@@ -38,6 +38,39 @@ var (
...
+	// CSI Controller resource requests/limits
+	// TODO: Reduce these values and reach optimistic values without effecting performance
+	TopolvmControllerMemRequest = "250Mi"
+	TopolvmControllerMemLimit   = "250Mi"
+	TopolvmControllerCPURequest = "250m"
+	TopolvmControllerCPULimit   = "250m"
+
+	TopolvmControllerCsiProvisionMemRequest = "250Mi"
+	TopolvmControllerCsiProvisionMemLimit   = "250Mi"
+	TopolvmControllerCsiProvisionCPURequest = "250m"
```
nbalacha, line 69: Lets reduce the cpu request and limit  to 100m and see if it works.

leelavg:
Thanks all for your early reviews, I'll address them at the earliest 😀

nbalacha:
Does the topolvm-controller need anything special wrt SCCs?

leelavg:
> Does the topolvm-controller need anything special wrt SCCs?

Afair, normal rbac would do, I will add them (rbac) based on topolvm helm chart.

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	volumes := []corev1.Volume{
+		{Name: "socket-dir", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+		{Name: "certs", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+	}
+
+	// TODO: Remove custom generation of TLS certs, find where it's being used in the first place in Topolvm Code
+	iContainers := []corev1.Container{*getInitContainer()}
```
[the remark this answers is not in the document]
leelavg: - ack, I wasn't able to confirm where does topolvm actually use these certs and I mean that in the comment.
- for now we are going with manual generation of certs using initContainer and replacing it with cert-manager is being investigated

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		"openssl req -nodes -x509 -newkey rsa:4096 -subj '/DC=self_signed_certificate' -keyout /certs/tls.key -out /certs/tls.crt -days 365",
+	}
+
+	volumeMounts := []corev1.VolumeMount{
+		{Name: "certs", MountPath: "/certs"},
+	}
+
+	ssCertGenerator := &corev1.Container{
+		Name:         "self-signed-cert-generator",
+		Image:        "alpine/openssl",
```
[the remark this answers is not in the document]
leelavg: - for now we just need to generate and use those certs, so changing base image would have any effect in openssl?

controllers/topolvm_csi_driver.go
```diff
@@ -0,0 +1,99 @@
...
+	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	cutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	driverName = "csi-driver"
```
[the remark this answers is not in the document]
leelavg: ack

controllers/topolvm_csi_driver.go
leelavg: - it's really beneficial in throw off containers and for some workloads which just needs storage to save logs which are of no use and need not to be persistent (let's say initialisation logs of an nginx)
- topolvm supports this with ease and tested, IMHO it's better to have this

controllers/topolvm_controller.go
leelavg: @nbalacha 

- I did some testing using configmaps with finalizers and ownerreferences and here's the gist in current context

1. If controller deployment were to depend on lvmcluster CR, if I set non-blocking ownerref on deployment and you set finalizer on CR, when you remove the finalizer then only controller deployment will be deleted
2. However the current reconciler only removes the finalizer when all the units get deleted, in this case I do not depend on lvm CR to be the owner and listen to reconciler to delete deployment whenever it's called

- I see two ways out of it
1. You do not change the reconcile and send delete op to the resource units and I'll add deletion logic
2. You can just remove the finalizer and send a delete op to the resource units and reconcile for nil (because some resources like csidriver is cluster scoped and doesn't respect ownerref)?

leelavg:
> Does the topolvm-controller need anything special wrt SCCs?

after referring to topolvm controller rbac, I don't see the need for SCC, those will be need for vgmanager and nodeplugin.

controllers/topolvm_csi_driver.go
nbalacha: Since we are not supporting this for 4.10, I think we should remove it. We can add it in a future release.

controllers/topolvm_csi_driver.go
leelavg: changing it to csiDriverResource to avoid name clashes.

leelavg: changing it to `csiDriverResource` to avoid name clashes.

controllers/topolvm_controller.go
nbalacha: Lets stick with the current approach for now - the ensureDeleted will effectively be a noop for topolvm-controller. The lvmcluster controller will remove the finalizer when all resource managers return nil . The LVMCluster object will then be removed followed the topolvm-controller.

controllers/lvmcluster_controller.go
```diff
@@ -95,7 +95,11 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 // errors returned by this will be updated in the reconcileSucceeded condition of the LVMCluster
 func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMCluster) (ctrl.Result, error) {
-	resourceList := []resourceManager{}
+	// TODO: Resource deletion should be in reverse order of install
+	resourceList := []resourceManager{
```
nbalacha, line 99: This will need to be split up amongst the 2 PRs (CSIDriver and topolvm-controller)

controllers/lvmcluster_controller.go
```diff
@@ -95,7 +95,11 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 // errors returned by this will be updated in the reconcileSucceeded condition of the LVMCluster
 func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMCluster) (ctrl.Result, error) {
-	resourceList := []resourceManager{}
+	// TODO: Resource deletion should be in reverse order of install
```
nbalacha, line 98: Not really necessary in this case.

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (c controllerPlugin) getName() string {
+	return controllerName
+}
+
+func (c controllerPlugin) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+
+	// get the desired state of topolvm controller deployment
+	controllerDeployment := getControllerDeployment(lvmCluster)
+	result, err := cutil.CreateOrUpdate(ctx, r.Client, controllerDeployment, func() error {
```
nbalacha, line 34: Better to update only if there is a change in the spec.

controllers/topolvm_controller.go
leelavg: - uh oh, just saw this review after adding tests 
- anyways will raise separate PRs

controllers/lvmcluster_controller_test.go
```diff
@@ -43,23 +50,79 @@ var _ = Describe("LVMCluster controller", func() {
...
 				if err != nil {
 					return false
 				}
-				return lvmCluster1.Status.Ready
+				return true
 			}, timeout, interval).Should(BeTrue())
-			// Let's make sure our Schedule string value was properly converted/handled.
-			Expect(lvmCluster1.Status.Ready).Should(Equal(true))
+
+			By("Confirming deletion of lvm cluster")
```
nbalacha, line 94: Please move the deletion to a separate test.

controllers/lvmcluster_controller_test.go
```diff
@@ -43,23 +50,79 @@ var _ = Describe("LVMCluster controller", func() {
...
+			// placeholder to check CR status.Ready field to be true
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, lvmClusterName, lvmClusterOut)
+				if err != nil {
+					return false
+				}
+				return lvmClusterOut.Status.Ready
+			}, timeout, interval).Should(Equal(true))
+
+			// present of csi driver resource
```
nbalacha, line 74: Nit: presence of csi driver resource

controllers/lvmcluster_controller_test.go
leelavg: - this is all part of `It("Reconciles an LvmCluster, "` if I were to move it to a separate test then I think I need to deploy again and then delete

controllers/topolvm_controller.go
leelavg: - Topolvm Controller deployment is resistant to LVM Cluster CR Updates
- After first creation the result will mostly be `Unchanged`
- Current implementation is used only to facilitate upgrade due to change in images, if both desired and existing state is same then it'll be `Unchanged`

controllers/topolvm_controller.go
leelavg: - I'm continuing with a non-blocking owner reference and also implementing custom logic for controller deletion
- We can decide what exactly to be used as it's not of utmost importance as of now

leelavg:
- this PR is split into #17, #18 and #21

PR #17, opened 2021-12-10
test: refactor and addition of CR deletion test

leelavg:
- use envtest for verifying reconciliation of lvm cluster CR

### DIFF
--- a/controllers/lvmcluster_controller_test.go
+++ b/controllers/lvmcluster_controller_test.go
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
 		interval = time.Millisecond * 250
 	)
 
-	Context("LvmCluster reconcile", func() {
-		It("Reconciles an LvmCluster, ", func() {
-			By("Indicate setting status to ready")
-			ctx := context.Background()
-			lvmCluster := &lvmv1alpha1.LVMCluster{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      testLvmClusterName,
-					Namespace: testLvmClusterNamespace,
-				},
-				Spec: lvmv1alpha1.LVMClusterSpec{
-					DeviceClasses: []lvmv1alpha1.DeviceClass{{Name: "test"}},
-				},
-			}
-			Expect(k8sClient.Create(ctx, lvmCluster)).Should(Succeed())
+	ctx := context.Background()
 
-			//Check that the status.Ready field has been set to true. This is a placeholder test and will
-			// be modified to check for the actual resources once they are implemented.
+	// LVMCluster CR details
+	lvmClusterName := types.NamespacedName{Name: testLvmClusterName, Namespace: testLvmClusterNamespace}
+	lvmClusterOut := &lvmv1alpha1.LVMCluster{}
+	lvmClusterIn := &lvmv1alpha1.LVMCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testLvmClusterName,
+			Namespace: testLvmClusterNamespace,
+		},
+		Spec: lvmv1alpha1.LVMClusterSpec{
+			DeviceClasses: []lvmv1alpha1.DeviceClass{{Name: "test"}},
+		},
+	}
 
-			lvmClusterLookupName := types.NamespacedName{Name: testLvmClusterName, Namespace: testLvmClusterNamespace}
-			lvmCluster1 := &lvmv1alpha1.LVMCluster{}
+	Context("Reconciliation on creating an LVMCluster CR", func() {
+		It("should reconcile LVMCluster CR creation, ", func() {
+			By("verifying CR status.Ready is set to true on reconciliation")
+			Expect(k8sClient.Create(ctx, lvmClusterIn)).Should(Succeed())
 
+			// placeholder to check CR status.Ready field to be true
 			Eventually(func() bool {
-				err := k8sClient.Get(ctx, lvmClusterLookupName, lvmCluster1)
+				err := k8sClient.Get(ctx, lvmClusterName, lvmClusterOut)
 				if err != nil {
 					return false
 				}
-				return lvmCluster1.Status.Ready
-			}, timeout, interval).Should(BeTrue())
-			// Let's make sure our Schedule string value was properly converted/handled.
-			Expect(lvmCluster1.Status.Ready).Should(Equal(true))
+				return lvmClusterOut.Status.Ready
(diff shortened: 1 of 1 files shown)


controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
+		},
+		Spec: lvmv1alpha1.LVMClusterSpec{
+			DeviceClasses: []lvmv1alpha1.DeviceClass{{Name: "test"}},
+		},
+	}
 
-			lvmClusterLookupName := types.NamespacedName{Name: testLvmClusterName, Namespace: testLvmClusterNamespace}
-			lvmCluster1 := &lvmv1alpha1.LVMCluster{}
+	Context("LVMCluster creation", func() {
+		It("should create an LVMCluster, ", func() {
```
nbalacha, line 49: The test should check that the reconcile succeeds on creation. It is not testing the LVMCluster creation itself.

controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
+			By("indicating CR status to be ready after CR is deployed")
```
nbalacha, line 50: "Setting status.Ready to true on reconciliation"

controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
-			}, timeout, interval).Should(BeTrue())
-			// Let's make sure our Schedule string value was properly converted/handled.
-			Expect(lvmCluster1.Status.Ready).Should(Equal(true))
+				return lvmClusterOut.Status.Ready
+			}, timeout, interval).Should(Equal(true))
 		})
+	})
+
+	Context("LVMCluster deletion", func() {
+		It("should delete an LVMCluster, ", func() {
```
nbalacha, line 65: It should delete the resources on LVMCluster deletion.

nbalacha (changes requested):

jmolmo:
IMHO we are not testing the functionality of the controller. I mean we are just checking that  [controller runtime client](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/client) works properly using a CRD. And that is something [that has been already verified in the client package](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/example_test.go ).

I think that it has more meaning and utility for us the test proposed in #13, where after the creation of the lvmcluster CRD we verified the existence of the CSIDriver resource and topolvm deployment.

In any case ... is difficult for me to see this kind of test as "unit test" ,,, is more like a mix between integration test (without CI env) and "extended" unit tests.

controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
-			// Let's make sure our Schedule string value was properly converted/handled.
-			Expect(lvmCluster1.Status.Ready).Should(Equal(true))
+				return lvmClusterOut.Status.Ready
+			}, timeout, interval).Should(Equal(true))
 		})
+	})
+
+	Context("Deletion of LVMCluster deployed resources", func() {
+		It("should check successful reconciliation of LVMCluster ", func() {
+			By("confirming absence of lvm cluster CR")
```
nbalacha, line 66: The reconciliation does not delete the lvmCluster CR. The test should check that the LvmCluster deletion removes the other resources created by the operator.

leelavg:
> And that is something that has been already verified in the client package
- It was verified for an example resource creation, in our case we want to test creation of our CR with default spec for now and check status is true & reconciliation started and working as expected 

> I think that it has more meaning and utility for us the test proposed in #13
- the said pr will be split into three and this is the first one where I'm trying to create a layout for e2e-ish kind of flow and rest of 2 prs will build up on this layout

> In any case ... is difficult for me to see this kind of test as "unit test" ,,, is more like a mix between integration test (without CI env) and "extended" unit tests.
- yep, it's arguable, in a unit test-ish setup we have test spies, mocks/fakes etc however I don't want to stop moving forward to decide on how unit tests should look like at this point/pace of development 😅

IMHO, if we had flow charts then it suites better to commit failing tests first and then code to fix those tests

I'm open to suggestions on how good to refactor tests or what should be tested as part of unit tests and e2e tests.

Pls correct me if I presented something not relatable/right 😀

controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
+			Namespace: testLvmClusterNamespace,
+		},
+		Spec: lvmv1alpha1.LVMClusterSpec{
+			DeviceClasses: []lvmv1alpha1.DeviceClass{{Name: "test"}},
+		},
+	}
 
-			lvmClusterLookupName := types.NamespacedName{Name: testLvmClusterName, Namespace: testLvmClusterNamespace}
-			lvmCluster1 := &lvmv1alpha1.LVMCluster{}
+	Context("LVMCluster reconciliation on installing CR", func() {
```
nbalacha, line 48: Reconciliation on creating an LVMCluster CR

controllers/lvmcluster_controller_test.go
```diff
@@ -30,38 +30,47 @@ var _ = Describe("LVMCluster controller", func() {
...
-				return lvmCluster1.Status.Ready
-			}, timeout, interval).Should(BeTrue())
-			// Let's make sure our Schedule string value was properly converted/handled.
-			Expect(lvmCluster1.Status.Ready).Should(Equal(true))
+				return lvmClusterOut.Status.Ready
+			}, timeout, interval).Should(Equal(true))
 		})
+	})
+
+	Context("LVMCluster reconciliation on uninstalling CR", func() {
```
nbalacha, line 64: "reconciliation on deleting the LVMCluster CR"

jmolmo (approved):
I wonder is we can use the structure of this test to do all the "unit test"  for all the "functional units".
We need always to create the lvm cluster and start the reconcile loop, after that ,, the verification we need to do for each unit are basically the same ( the unit is created, deployed, and it is composed by several containers) .

leelavg:
@jmolmo yes, that's the plan. 

The current pr creates a basic skeleton and individual component owners add their units with their implementation.

nbalacha (approved):

PR #19, opened 2021-12-13
must-gather: initial commit

nbalacha:
The lvm must-gather will capture lvm related information
from an OCP cluster.

### DIFF
--- a/must-gather/README.md
+++ b/must-gather/README.md
@@ -1,0 +1,37 @@
+OCS must-gather
+=================
+
+`lvm-must-gather` is a tool built on top of [OpenShift must-gather](https://github.com/openshift/must-gather)
+that expands its capabilities to gather LVM Operator information.
+
+### Usage
+```sh
+oc adm must-gather --image=quay.io/ocs-dev/lvm-must-gather
+```
+
+The command above will create a local directory with a dump of the lvm state.
+
+You will get a dump of:
+- The LVM Operator namespaces (and its children objects)
+- All namespaces (and their children objects) that belong to any LVM resources
+- All LVM CRD's definitions
+
+
+### How to Contribute
+
+#### Contribution Flow
+Developers must follow these steps to make a change:
+1. Fork the `red-hat-storage/lvm-operator` repository on GitHub.
+2. Create a branch from the master branch, or from a versioned branch (such
+   as `release-4.2`) if you are proposing a backport.
+3. Make changes.
+4. Ensure your commit messages include sign-off.
+5. Push your changes to a branch in your fork of the repository.
+6. Submit a pull request to the `red-hat-storage/lvm-operator` repository.
+7. Work with the community to make any necessary changes through the code
+   review process (effectively repeating steps 3-7 as needed).
+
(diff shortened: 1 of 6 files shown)


must-gather/Centos-Base.repo
```diff
@@ -0,0 +1,6 @@
+[centos]
+name=CentOS-7
```
jmolmo, line 2: This tool is completely linked to RH. Why not to use an UBI image?

must-gather/README.md
```diff
@@ -0,0 +1,37 @@
+OCS must-gather
+=================
+
+`lvm-must-gather` is a tool built on top of [OpenShift must-gather](https://github.com/openshift/must-gather)
```
jmolmo, line 4: The name of this tool is OCS must-gather or "lvm must gather"?

jmolmo (commented):
If we are going to use this tool from inside topolvm, I think we need a better way to retrieve the information directly from the "code", I mean.. we will need to do several changes  to use the tool as library.
In other side, it is for openshift... now we have a very good opportunity to make it more generic, allowing to be executed in pure k8s environments. 
I also have doubts if the topolvm operator repository is the right place to develop it, I think the tool  has enough utility and importance to have  his own repository, and to work as a separate project.

Can you clarify what is going to be the roadmap for this tool?
Why do you think we need to have it inside topolvm ( as a separate tool that dumps results in a directory) ?

nbalacha:
> If we are going to use this tool from inside topolvm, I think we need a better way to retrieve the information directly from the "code", I mean.. we will need to do several changes to use the tool as library. In other side, it is for openshift... now we have a very good opportunity to make it more generic, allowing to be executed in pure k8s environments. I also have doubts if the topolvm operator repository is the right place to develop it, I think the tool has enough utility and importance to have his own repository, and to work as a separate project.
> 
> Can you clarify what is going to be the roadmap for this tool? Why do you think we need to have it inside topolvm ( as a separate tool that dumps results in a directory) ?


Hi Juan,

The must-gather command is specific to Openshift and required for debugging. 
More information about it is available at : https://docs.openshift.com/container-platform/4.4/support/gathering-cluster-data.html

Each openshift operator provides a must-gather utility which is currently implemented in each operator's code base. There have been discussions about moving them all to separate must-gather repo, but until that is done we will have it as part of lvm-operator. We will be building and shipping a separate lvm-must-gather image.

must-gather/README.md
nbalacha: This will be lvm-must-gather

jmolmo:
> The must-gather command is specific to Openshift and required for debugging.

Thanks for the clarification @nbalacha , lot of things are completely new for me.. sorry for the dumb questions

jmolmo (approved):

must-gather/Centos-Base.repo
nbalacha: Looks like this is not required at all.  I will remove it.

nbalacha:
Not dumb at all. A lot of this is new to many of us.

leelavg (approved):
/lgtm

PR #20, opened 2021-12-13
controller: sets namespace to openshift-storage

nbalacha:
Changed the default operator namespace to openshift-storage.


jmolmo (approved):
LGTM. But we will need "a list of things to change" if we want the operator will be under the topolvm umbrella.

leelavg:
@nbalacha i think this can be merged?

PR #21, opened 2021-12-13
feat: implementation of topolvm controller resource unit

leelavg:
- implemented create and delete reconcile actions for TopolvmController
- added rbac roles to manager for manage TopolvmController resource
- added rbac roles for topolvm-controller to manage CSI Controller functionanity

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -6,7 +6,7 @@ namespace: lvm-operator-system
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: lvm-operator-
+# namePrefix: lvm-operator-
 
 # Labels to add to all resources and selectors.
 #commonLabels:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,17 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
   - lvm.topolvm.io
   resources:
   - lvmclusters
--- a/config/rbac/topolvm_controller_role.yaml
+++ b/config/rbac/topolvm_controller_role.yaml
@@ -1,0 +1,271 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: topolvm-controller
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - delete
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  - csidrivers
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - topolvm.cybozu.com
+  resources:
+  - logicalvolumes
+  - logicalvolumes/status
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-csi-provisioner
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumes
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - list
+  - watch
+  - create
+  - update
+  - patch
+- apiGroups:
+  - snapshot.storage.k8s.io
+  resources:
+  - volumesnapshots
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - snapshot.storage.k8s.io
+  resources:
+  - volumesnapshotcontents
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - csinodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - volumeattachments
+  verbs:
+  - get
+  - list
+  - watch
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-csi-resizer
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumes
+  verbs:
+  - get
+  - list
+  - watch
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - list
+  - watch
+  - create
+  - update
+  - patch
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-leader-election
+rules:
+- apiGroups:
+    - ""
+    - coordination.k8s.io
+  resources:
+    - configmaps
+    - leases
+  verbs:
+    - get
+    - list
+    - watch
+    - create
+    - update
+    - patch
+    - delete
+- apiGroups:
+    - ""
+  resources:
+    - events
+  verbs:
+    - create
+    - patch
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-csi-provisioner
+rules:
+- apiGroups:
+    - coordination.k8s.io
+  resources:
+    - leases
+  verbs:
+    - get
+    - watch
+    - list
+    - delete
+    - update
+    - create
+- apiGroups:
+    - storage.k8s.io
+  resources:
+    - csistoragecapacities
+  verbs:
+    - get
+    - list
+    - watch
+    - create
+    - update
+    - patch
+    - delete
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - get
+- apiGroups:
+    - apps
+  resources:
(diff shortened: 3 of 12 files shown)


nbalacha (changes requested):
Please split this PR into multiple commits. (code + manifests)

config/rbac/topolvm_controller_role_bindings.yaml
```diff
@@ -0,0 +1,79 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: topolvm-system:controller
```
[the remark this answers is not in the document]
leelavg: - yes, it is and does work. But I'm renaming it.
- I wasn't able to find much documentation about that and at the same time it seem to signify namespace in which the resource using this bindings is deployed

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,279 @@
...
+type controllerPlugin struct{}
+
+// controllerPlugin unit satisfies resourceManager interface
+var _ resourceManager = controllerPlugin{}
+
+func (c controllerPlugin) getName() string {
+	return controllerName
+}
+
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;update;delete;get;list;watch
```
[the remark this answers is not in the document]
leelavg: - operator-sdk seems to be building up sharedinformercache using client-go's watch&list functions and so added those verbs

config/default/kustomization.yaml
```diff
@@ -6,7 +6,7 @@ namespace: lvm-operator-system
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: lvm-operator-
+# namePrefix: lvm-operator-
```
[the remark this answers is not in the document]
leelavg: the issue here is, every resource name will have this prefix and it creates trouble while using topolvmcontroller service account name.

we can get away either not using one or use a consistent namePrefix.

leelavg:
- tested against a live cluster by deploying, updating and deleting lvmcluster CR and observing reconciliation of topolvm controller resource unit
- deployed csi driver manually as that's part of another PR and this resource depends on csidriver presence

config/default/kustomization.yaml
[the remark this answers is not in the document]
leelavg: as discussed, not using the prefix.

config/rbac/topolvm_controller_role.yaml
```diff
@@ -0,0 +1,271 @@
...
+    - apps
+  resources:
+    - replicasets
+  verbs:
+    - get
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-resizer-cfg
```
nbalacha, line 258: Why call it "external-resizer-cfg"  instead of "external-resizer"?

config/rbac/topolvm_controller_role.yaml
```diff
@@ -0,0 +1,271 @@
...
+  - volumeattachments
+  verbs:
+  - get
+  - list
+  - watch
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-external-resizer-runner
```
nbalacha, line 143: Please remove "runner" from the name.

config/rbac/topolvm_controller_role.yaml
```diff
@@ -0,0 +1,271 @@
...
+  resources:
+    - events
+  verbs:
+    - create
+    - patch
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-provisioner-cfg
```
nbalacha, line 217: Please remove "cfg" from the name

nbalacha (commented):
The commit msg for the first commit refers to changes made in commit2 . Please correct the commit msgs.

config/rbac/topolvm_controller_role_bindings.yaml
```diff
@@ -0,0 +1,79 @@
...
+  name: topolvm-external-provisioner-runner
+  apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topolvm-csi-resizer-role
+subjects:
+  - kind: ServiceAccount
+    namespace: lvm-operator-system
```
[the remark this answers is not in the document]
leelavg: - No, it doesn't have to be, all it does it adds `namespace` mentioned in `default/kustomization.yaml`

controllers/defaults.go
```diff
@@ -38,6 +38,41 @@ var (
...
+
+	TopolvmCsiProvisionerMemRequest = "100Mi"
+	TopolvmCsiProvisionerMemLimit   = "100Mi"
+	TopolvmCsiProvisionerCPURequest = "100m"
+	TopolvmCsiProvisionerCPULimit   = "100m"
+
+	// CSI Provisioner requires below environment values to make use of CSIStorageCapacity
+	PodNameSpaceEnv = "POD_NAMESPACE"
+	PodNameEnv      = "POD_NAME"
+	NameSpaceEnv    = "NAMESPACE"
```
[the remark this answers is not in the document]
leelavg: - Only `NAMESPACE` would suffice

config/rbac/role.yaml
```diff
@@ -6,6 +6,17 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
+- apiGroups:
```
jmolmo, line 9: Why do we need two files (role.yaml and topolvm_controller_role.yaml) for defining roles/cluster roles?

leelavg: in rbac folder whatever isn't prefixed with topolvm corresponds to lvm operator rbac

For now, components will add their rbac and towards the end (after all units implementation) if anything is sharing the same roles those can be clubbed, operator rbac be renamed to correctly reflect their usage.

jmolmo (approved):

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,325 @@
...
+		{Name: "socket-dir", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+		{Name: "certs", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+	}
+
+	// TODO: Remove custom generation of TLS certs, current it's being used in topolvm controller manager
+	initContainers := []corev1.Container{
+		*getInitContainer(),
+	}
+
+	// getall containers that are part of csi controller deployment
```
sp98, line 103:
```suggestion
	// get all containers that are part of csi controller deployment
```

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;update;delete;get;list;watch
+
+func (c topolvmController) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+
+	// get the desired state of topolvm controller deployment
+	controllerDeployment := getControllerDeployment(lvmCluster)
+
+	result, err := cutil.CreateOrUpdate(ctx, r.Client, controllerDeployment, func() error { return nil })
```
sp98, line 38: The `MutateFn` (last parameter of this function) is empty.  So it will never update the deployment. I don't think that's what we want here.

sp98 (changes requested):

controllers/topolvm_controller.go
sp98: or maybe I'm missing something here.

leelavg: > it will never update the deployment
- I don't think so, we don't want to mutate the deployment, if there's an upgrade scenario then controller-util will compare the spec's and update it (if you meant deployment update)
- When LVMCluster CR is changed then also deployment doesn't need to be updated (if you meant in CR context)

controllers/topolvm_controller.go
leelavg: ack, can I take this as part of separate PR?

leelavg: ack, can I take this as part of separate PR?

sp98: sure. Separate PR sounds good.

controllers/topolvm_controller.go
sp98: thanks for the clarification.

nbalacha (changes requested):
I see this error when I run `make deploy` after `make deploy`.

...
```
Error from server (NotFound): error when deleting "STDIN": configmaps "manager-config" not found
Error from server (NotFound): error when deleting "STDIN": services "controller-manager-metrics-service" not found
Error from server (NotFound): error when deleting "STDIN": deployments.apps "controller-manager" not found
```

leelavg:
>  I run `make deploy` after `make deploy`.
- I take that as `make undeploy` after `make deploy`
- Did you see below towards the end when you run `make deploy`
```
configmap/manager-config created
service/controller-manager-metrics-service created
deployment.apps/controller-manager created
```
- Pls confirm presence of above resources before you run `undeploy` that'd suffice
- I tested atleast twice on OCP (3 master + 3 worker) and sometimes `undeploy` says deployment is not found cause namespace deletion triggers deployment deletion as well
- And there's a chance that by the time above services are called for deletion by `undeploy` they were already deleted

nbalacha:
> > I run `make deploy` after `make deploy`.
> 
> * I take that as `make undeploy` after `make deploy`
> * Did you see below towards the end when you run `make deploy`
> 
> ```
> configmap/manager-config created
> service/controller-manager-metrics-service created
> deployment.apps/controller-manager created
> ```
> 
> * Pls confirm presence of above resources before you run `undeploy` that'd suffice
> * I tested atleast twice on OCP (3 master + 3 worker) and sometimes `undeploy` says deployment is not found cause namespace deletion triggers deployment deletion as well
> * And there's a chance that by the time above services are called for deletion by `undeploy` they were already deleted

I'll check but this error is not seen without this PR.

leelavg:
> I'll check but this error is not seen without this PR.
- ack, it's not seen in other PRs because there wasn't much roles/bindings/sa etc and above race will not happen
- either ways, will await your confirmation

controllers/topolvm_controller.go
```diff
@@ -0,0 +1,325 @@
...
+		r.Log.Info("csi controller", "operation", result, "name", controllerDeployment.Name)
+	case cutil.OperationResultUpdated:
+		r.Log.Info("csi controller", "operation", result, "name", controllerDeployment.Name)
+	case cutil.OperationResultNone:
+		r.Log.Info("csi controller", "operation", result, "name", controllerDeployment.Name)
+	default:
+		r.Log.Error(err, "csi controller reconcile failure", "name", controllerDeployment.Name)
+		return err
+	}
+
```
nbalacha, line 52: Please replace this with 
	if err != nil {
		r.Log.Error(err, "reconcile failure", "name", controllerDeployment.Name)
	} else {
		r.Log.Info("csi controller", "operation", result, "name", controllerDeployment.Name)
	}

controllers/topolvm_controller.go
leelavg: done

nbalacha (changes requested):
Sorry. Did not notice this in earlier reviews. Everything else looks fine though we should probably validate the rbacs later.

leelavg:
> Sorry. Did not notice this in earlier reviews. Everything else looks fine though we should probably validate the rbacs later.

- my mistake too in missing this, thanks for finding it.
- validation of rbac wrt topolvm controller isn't needed, we should use consistent names for rbac used for lvm-operator after merging all components

nbalacha:
> > Sorry. Did not notice this in earlier reviews. Everything else looks fine though we should probably validate the rbacs later.
> 
> * my mistake too in missing this, thanks for finding it.
> * validation of rbac wrt topolvm controller isn't needed, we should use consistent names for rbac used for lvm-operator after merging all components

nbalacha:
Closed this by mistake. Apologies.

nbalacha (approved):
/lgtm

PR #24, opened 2021-12-15
ci: github action to run 'make test' on every pr

leelavg:

### DIFF
--- a/.github/workflows/on-pr-submit.yaml
+++ b/.github/workflows/on-pr-submit.yaml
@@ -1,0 +1,39 @@
+---
+name: test
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - 'doc/**'
+
+jobs:
+  test:
+    name: test
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pr-submit.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - 'doc/**'
+
+jobs:
+  envtest:
+    name: envtest
```
nbalacha, line 13: please rename this to  'test" both here and in the job name

leelavg:
> Please replace envtest there as well.

- I did see that actually and at the same time, current action has only envtest so didn't change that for easy ref
- Still want me to change?

nbalacha:
> > Please replace envtest there as well.
> 
> * I did see that actually and at the same time, current action has only envtest so didn't change that for easy ref
> * Still want me to change?

Yes.

jmolmo (approved):

nbalacha (approved):
/lgtm

PR #28, opened 2021-12-20
controller: operator now watches own namespace

nbalacha:
This introduces the WATCH_NAMESPACE env which is set the the operator's
own namespace. This allows the operator to be configured to watch only its
own namespace.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -52,5 +52,10 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
+        - name: WATCH_NAMESPACE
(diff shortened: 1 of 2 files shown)


config/manager/manager.yaml
```diff
@@ -52,5 +52,10 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
```
leelavg, line 55: - when installing via bundle we'll specify the namespace in which this operator should be installed?

nbalacha: yes

leelavg (approved):
/lgtm

PR #34, opened 2021-12-20
feat: add mutate function

jmolmo:
Implement the mutate function content for the createorUpdate topolvmdNode
resource.

### DIFF
--- a/controllers/topolvm_node.go
+++ b/controllers/topolvm_node.go
@@ -24,7 +24,6 @@ import (
 
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
-	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


controllers/topolvm_node.go
```diff
@@ -152,27 +190,25 @@ func getNodeDaemonSet(lvmCluster *lvmv1alpha1.LVMCluster) *v1.DaemonSet {
 	if tolerations != nil {
 		topolvmNodeTolerations = tolerations
 	}
-
+	labels := map[string]string{
+		"app":                   topolvmNodeName,
+		"topolvm.io/lvmcluster": lvmCluster.Name,
```
leelavg, line 195: - there's a possibility that this might become ambiguous as there are recent discussions to supports multiple LVM CRs
- let's consider three nodes, out of which create daemonset targeted a single node initially and later another CR targeted remaining nodes, in that scenario this label will represent new CR only
- unless there's specific use IMHO it's better not to use this specific label

jmolmo: In the case you expose, what we will  have is a couple of CRs with the same labels. Is clear that, we will need to implement the management of several CRs, but i cannot see the problem of having the same labels in two CRs. Can you remark what is the problem you see?

leelavg: Ack, I'm concerned about `topolvm.io/lvmcluster` value as this'll be overwritten in all node daemonsets if there's a new CR.

leelavg:
Forgot to mention earlier, we need to assign containers spec as well in mutate function.

controllers/topolvm_node.go
```diff
@@ -101,7 +140,7 @@ func extractNodeSelectorAndTolerations(lvmCluster lvmv1alpha1.LVMCluster) (*core
 	return nil, nil
 }
 
-func getNodeDaemonSet(lvmCluster *lvmv1alpha1.LVMCluster) *v1.DaemonSet {
+func getNodeDaemonSet(lvmCluster *lvmv1alpha1.LVMCluster, nodeSelector *corev1.NodeSelector, tolerations []corev1.Toleration) *v1.DaemonSet {
```
nbalacha, line 143: It is not required to pass it in - the extract function can be called inside this function

controllers/topolvm_node.go
nbalacha: We don't need this label.

jmolmo: I will remove

sp98 (approved):

controllers/topolvm_node.go
```diff
@@ -47,14 +46,45 @@ func (n topolvmNode) getName() string {
...
+		// For topolvm Node, we have containers, node selector and toleration terms
+
+		// containers
+		ds.Spec.Template.Spec.Containers = dsTemplate.Spec.Template.Spec.Containers
+
+		// tolerations
+		ds.Spec.Template.Spec.Tolerations = dsTemplate.Spec.Template.Spec.Tolerations
+
+		// nodeSelector if non-nil
+		if dsTemplate.Spec.Template.Spec.Affinity != nil {
```
nbalacha, line 77: This needs to updated even if it is non-nil. We can change the nodeselector.

nbalacha: ignore this comment.

PR #35, opened 2021-12-20
feat: Use custom SCC for topolvm-node

jmolmo:
Replaced the <privileged> SCC with the custom topolvm-node SCC


config/rbac/topolvm_node_scc.yaml
```diff
@@ -0,0 +1,30 @@
...
+  type: RunAsAny
+fsGroup:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+volumes:
+  - 'configMap'
+  - 'emptyDir'
+  - 'secret'
+  - 'hostPath'
```
sp98, line 30: How is this SCC associated with the topolvm-node Service account?  There is a `user` filed that contains the list of serviceAccounts. For example- [rook](https://github.com/rook/rook/blob/fffb86295600196389de81905c3402f9aec6e751/deploy/examples/operator-openshift.yaml#L42)

config/rbac/topolvm_node_scc.yaml
```diff
@@ -0,0 +1,30 @@
...
+        features and the ability to run as any user, any group, any fsGroup, and with
+        any SELinux context.  WARNING: this is the most relaxed SCC and should be used
+        only for cluster administration. Grant with caution.'
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: true
+allowHostPorts: false
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: true
```
leelavg, line 16: - not immediate but our final goal should be to use specific capabilities that the container needs and curtail everything else.

leelavg (approved):
/lgtm

config/rbac/topolvm_node_scc.yaml
leelavg: - thanks for finding this, I saw `users` field somewhere and forgot where that was during our meetings
- in current scenario we are using ClusterRole & ClusterRoleBinding to use this SCC
- in merged commit it was using `privileged` directly and this changes that to use `topolvm-node` named SCC

config/rbac/topolvm_node_scc.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: topolvm-node
+  annotations:
+    kubernetes.io/description: 'privileged allows access to all privileged and host
+        features and the ability to run as any user, any group, any fsGroup, and with
+        any SELinux context.  WARNING: this is the most relaxed SCC and should be used
+        only for cluster administration. Grant with caution.'
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: true
```
leseb, line 13: Why?

leseb (commented):
Generally it's good to comment on why we require and permission, when we hope to remove it (if ever possible).

config/rbac/topolvm_node_scc.yaml
jmolmo: @sp98: I have include the different settings reading [what they do ](https://docs.openshift.com/container-platform/4.9/rest_api/security_apis/securitycontextconstraints-security-openshift-io-v1.html). 
As @leelavg commented we are using a cluster role binding to to associate the sa with the scc. Include the user in the scc seems to link the scc to a single namespace (because the way you need to specify the user), which i think is very limited for our needs.

config/rbac/topolvm_node_scc.yaml
jmolmo: https://github.com/red-hat-storage/lvm-operator/blob/848792e3cda401d0567e0140d006b068e3c54adf/controllers/topolvm_node.go#L182

leseb: That does not tell why we need `HostPID`.

config/rbac/topolvm_node_scc.yaml
```diff
@@ -0,0 +1,30 @@
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: topolvm-node
+  annotations:
+    kubernetes.io/description: 'privileged allows access to all privileged and host
```
nbalacha, line 6: The description needs to be updated.

config/rbac/topolvm_node_scc.yaml
nbalacha: The operator only needs the permissions for the sa in the operator namespace as that is where it will create the containers. Can we try this with a Role?

config/rbac/topolvm_node_scc.yaml
```diff
@@ -0,0 +1,30 @@
...
+        features and the ability to run as any user, any group, any fsGroup, and with
+        any SELinux context.  WARNING: this is the most relaxed SCC and should be used
+        only for cluster administration. Grant with caution.'
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: true
+allowHostPorts: false
+allowPrivilegeEscalation: true
```
nbalacha, line 15: I don't think we need privilegeEscalation

nbalacha (changes requested):

config/rbac/topolvm_node_scc.yaml
leelavg: @leseb generally nodeplugin doesn't need it however lvmd in the daemonset needs this.

It uses `nsenter` to get into host pid nasemspace and run lvm ops

leseb: I don't think you need `HostPID` for `nsenter` to run any commands on the host. As far as I can tell, Rook uses nsenter already to exec command on the host and we don't use `HostPID`.

However, if you really feel like we do then let's comment on why. Thanks

jmolmo: Extract from [Openshift documentation](https://docs.openshift.com/container-platform/4.9/rest_api/security_apis/securitycontextconstraints-security-openshift-io-v1.html#specification):
```
allowHostPID | boolean | AllowHostPID determines if the policy allows host pid in the containers.
-- | -- | --
```
If you are asking why do we need hostpid: true in the pod, this is because lvmd needs to see the pid:1 on the host. ([here you can follow the complete discussion that drove to this point](https://github.com/topolvm/topolvm/issues/139#issuecomment-684069197))

But, for me is new this world of SCCs.. how do you suggest to add this requirement (see pid 1 in the host) in the SCC?

leseb: I don't mean to rehearse or argue on the past decision that led to enabling `HostPID`. I just want to make sure everybody knows and understands why it is needed. That's all :) So in the `config/rbac/topolvm_node_scc.yaml` I'm just suggesting adding more comments to explain why each option we enable is needed.

config/rbac/topolvm_node_scc.yaml
jmolmo: @nbalacha : We cannot use  role instead of cluster role, because we need permissions for cluster-scoped resources (like nodes).

jmolmo:
@nbalacha : We cannot use Role instead of Cluster Role, because we need permissions for cluster-scoped resources (like nodes).

config/rbac/topolvm_node_scc.yaml
jmolmo: it is needed because we use "privileged" containers:  if we change this setting , we have errors deploying the daemonset, example:

` Warning  FailedCreate  44s               daemonset-controller  Error creating: Pod "topolvm-node-qx67b" is invalid: [spec.containers[0].securityContext: Invalid value: core.SecurityContext{Capabilities:(*core.Capabilities)(nil), Privileged:(*bool)(0xc0571abd0a), SELinuxOptions:(*core.SELinuxOptions)(nil), WindowsOptions:(*core.WindowsSecurityContextOptions)(nil), RunAsUser:(*int64)(0xc0571abd60), RunAsGroup:(*int64)(nil), RunAsNonRoot:(*bool)(nil), ReadOnlyRootFilesystem:(*bool)(nil), AllowPrivilegeEscalation:(*bool)(0xc0571abc48), ProcMount:(*core.ProcMountType)(nil), SeccompProfile:(*core.SeccompProfile)(nil)}: cannot set `allowPrivilegeEscalation` to false and `privileged` to true`

nbalacha: ack.

config/rbac/topolvm_node_scc.yaml
jmolmo: @nbalacha : We cannot use Role instead of Cluster Role, because we need permissions for cluster-scoped resources (like nodes).

jmolmo:
I have extracted the topolvm-node scc from the cluster role, and created a specific cluster role and binding for the scc. In this way deployment can be more flexible. I guess that we also will need in a near future to add  PSPs, :-)

nbalacha (approved):

PR #38, opened 2021-12-22
feat: topolvm storage class resource unit implementation

leelavg:
- Implementation of resourceManager interface for topolvm storage class
- Addition of RBAC for operator to manage topolvm storage classes
- Envtest to verify creation and deletion of storage class based on CR

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -65,3 +65,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
--- a/controllers/topolvm_storageclass.go
+++ b/controllers/topolvm_storageclass.go
@@ -1,0 +1,112 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	cutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	scName = "topolvm-storageclass"
+)
+
+type topolvmStorageClass struct{}
+
+// topolvmStorageClass unit satisfies resourceManager interface
+var _ resourceManager = topolvmStorageClass{}
+
+func (s topolvmStorageClass) getName() string {
+	return scName
+}
+
+//+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;create;delete;watch;list
+
+func (s topolvmStorageClass) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+
+	// one storage class for every deviceClass based on CR is created
+	topolvmStorageClasses := getTopolvmStorageClasses(lvmCluster)
+	for _, sc := range topolvmStorageClasses {
+
+		// we anticipate no edits to storage class
+		result, err := cutil.CreateOrUpdate(ctx, r.Client, sc, func() error { return nil })
+		if err != nil {
+			r.Log.Error(err, "topolvm storage class reconcile failure", "name", sc.Name)
+			return err
+		} else {
+			r.Log.Info("topolvm storage class", "operation", result, "name", sc.Name)
+		}
+	}
+	return nil
+}
+
+func (s topolvmStorageClass) ensureDeleted(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+
+	// construct name of storage class based on CR spec deviceClass field and
+	// delete the corresponding storage class
+	for _, deviceClass := range lvmCluster.Spec.DeviceClasses {
+		sc := &storagev1.StorageClass{}
+		scName := fmt.Sprintf("topolvm-%s", deviceClass.Name)
+		err := r.Client.Get(ctx, types.NamespacedName{Name: scName}, sc)
+
+		if err != nil {
+			// already deleted in previous reconcile
+			if errors.IsNotFound(err) {
+				r.Log.Info("topolvm storage class is deleted", "StorageClass", scName)
+				return nil
+			}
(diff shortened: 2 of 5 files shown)


leelavg:
**Pending: It'll be part of separate PR**
- Based on lvmCluster CR, storage class unit also extracts tolerations
- It'll check the status of daemonset with these tolerations and if they are running then volume group creation was successful
- Then storageclass will respect reconciler `ensureCreated` and creates StorageClass or else it'll wait for next cycle.

controllers/topolvm_storageclass.go
```diff
@@ -0,0 +1,110 @@
...
+				Annotations: map[string]string{
+					"description": "Provides RWO and RWOP Filesystem & Block volumes",
+				},
+			},
+			Provisioner:          TopolvmCSIDriverName,
+			VolumeBindingMode:    &volumeBindingMode,
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters: map[string]string{
+				fmt.Sprintf("%s/device-class", TopolvmCSIDriverName): deviceClass.Name,
+				"csi.storage.k8s.io/fstype":                          "xfs",
```
nbalacha, line 104: Minor comment .Please move xfs to a default var.

controllers/topolvm_storageclass.go
```diff
@@ -0,0 +1,110 @@
...
+	// delete the corresponding storage class
+	for _, deviceClass := range lvmCluster.Spec.DeviceClasses {
+		sc := &storagev1.StorageClass{}
+		scName := fmt.Sprintf("topolvm-%s", deviceClass.Name)
+		err := r.Client.Get(ctx, types.NamespacedName{Name: scName}, sc)
+
+		if err != nil {
+			// already deleted in previous reconcile
+			if errors.IsNotFound(err) {
+				r.Log.Info("topolvm storage class", "StorageClass", scName)
```
jmolmo, line 60: Message is not clear... better something like:
"topolvm storage class already deleted"

controllers/topolvm_storageclass.go
```diff
@@ -0,0 +1,110 @@
...
+			r.Log.Error(err, "failed to retrieve topolvm storage class", "StorageClass", scName)
+			return err
+		}
+
+		// storageClass exists, initiate deletion
+		if sc.GetDeletionTimestamp().IsZero() {
+			if err = r.Client.Delete(ctx, sc); err != nil {
+				r.Log.Error(err, "failed to delete topolvm storage class", "StorageClass", scName)
+				return err
+			}
```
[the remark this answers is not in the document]
leelavg: above means we are unable to delete storage class or did I miss anything?

controllers/topolvm_storageclass.go
```diff
@@ -0,0 +1,110 @@
...
+
+func getTopolvmStorageClasses(lvmCluster *lvmv1alpha1.LVMCluster) []*storagev1.StorageClass {
+	sc := []*storagev1.StorageClass{}
+	allowVolumeExpansion := true
+	volumeBindingMode := storagev1.VolumeBindingWaitForFirstConsumer
+
+	for _, deviceClass := range lvmCluster.Spec.DeviceClasses {
+		storageClass := &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("topolvm-%s", deviceClass.Name),
```
sp98, line 94: would this ensure correct [naming conventions](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)? Can device class have upper case alphanumeric chars?

controllers/topolvm_storageclass.go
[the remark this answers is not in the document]
jmolmo: I mean.. just inform when the sc is deleted: 
```
		if sc.GetDeletionTimestamp().IsZero() {
			if err = r.Client.Delete(ctx, sc); err != nil {
				r.Log.Error(err, "failed to delete topolvm storage class", "StorageClass", scName)
				return err
			} else {
				r.Log.Info(err, "topolvm storage class succesfully deleted", "StorageClass", scName)
			}
```

controllers/topolvm_storageclass.go
```diff
@@ -0,0 +1,110 @@
...
+				Annotations: map[string]string{
+					"description": "Provides RWO and RWOP Filesystem & Block volumes",
```
sp98, line 96: are we supporting RWOP?

config/rbac/role.yaml
```diff
@@ -65,3 +65,13 @@ rules:
...
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
```
sp98, line 77: where do we need `watch`?

controllers/topolvm_storageclass.go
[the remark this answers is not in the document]
leelavg: it's already logged above and this codepath will not be touched in that case

jmolmo (changes requested):
I think it is OK. LGTM after make just a couple of changes in the log messages. I would like to comment that the name of the resource is "singular"  topolvmStorageClass, but what we are managing are several storageclasses. probably a name like "topolvmStorageClasses" would be more meaningful, but it is just an opinion. It is is ok to keep the name in singular.

config/rbac/role.yaml
leelavg: - it's probably for operator-sdk/controller-runtime to build sharedinformercache, I didn't dig into code though where that was defined
- when `watch` wasn't provided I saw errors logged in

controllers/topolvm_storageclass.go
leelavg: - either ways, there isn't any watch to limit access mode and I thought it's better to annotate the capabilities that the provisioner provides

controllers/topolvm_storageclass.go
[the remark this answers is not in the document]
jmolmo: you mean in a second pass?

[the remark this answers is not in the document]
leelavg: ack, I think it doesn't harm to log deletion is initiated.

controllers/topolvm_storageclass.go
leelavg: - good catch, do not want to implement it here, will track this as an issue. Sounds good?
- it's better to restrict deviceClass to lowercase but let's think about that later.

sp98: sounds good to me.

sp98 (approved):

jmolmo (approved):

nbalacha (approved):

PR #45, opened 2021-12-24
fix: use single image for operator and vgmanager

leelavg:


controllers/vgmanager_daemonset.go
```diff
@@ -189,3 +207,31 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
 	setDaemonsetNodeSelector(nodeSelector, &ds)
 	return ds
 }
+
+func getRunningPodImage(r *LVMClusterReconciler, ctx context.Context) (string, error) {
```
sp98, line 211: how about adding the image as env in the csv?

leelavg: - that's also and option, we need to update both image and env variable in that case.

controllers/vgmanager_daemonset.go
```diff
@@ -189,3 +207,31 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
...
+
+	// 'POD_NAME' and 'WATCH_NAMESPACE' are set in env of lvm-operator when running as a container
+	podName := os.Getenv("POD_NAME")
+	if podName == "" {
+		return "", fmt.Errorf("failed to get image from running pod")
+	}
+
+	podNamespace := os.Getenv("WATCH_NAMESPACE")
+	if podNamespace == "" {
+		return "", fmt.Errorf("failed to get naemspace from running pod")
```
nbalacha, line 221: Nit: typo : namespace

controllers/vgmanager_daemonset.go
```diff
@@ -112,17 +117,30 @@ var (
...
 	privileged := true
 	var zero int64 = 0
+
+	// try to get running pod image, on failure fallback to default image
+	image, err := getRunningPodImage(r, ctx)
+	if err != nil {
+		r.Log.Info(err.Error())
+		image = VGManagerImage
+	}
+	r.Log.Info(fmt.Sprintf("using image '%s' for VGManager", image))
```
nbalacha, line 134: Please use structured logging:
r.Log.Info("creating VG manager deployment", "image", image)

config/manager/manager.env
```diff
@@ -4,4 +4,4 @@ CSI_REGISTRAR_IMAGE=k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
 CSI_PROVISIONER_IMAGE=k8s.gcr.io/sig-storage/csi-provisioner:v3.0.0
 CSI_LIVENESSPROBE_IMAGE=k8s.gcr.io/sig-storage/livenessprobe:v2.5.0
 CSI_RESIZER_IMAGE=k8s.gcr.io/sig-storage/csi-resizer:v1.3.0
-VGMANAGER_IMAGE=quay.io/ocs-dev/vgmanager:latest
+VGMANAGER_IMAGE=quay.io/ocs-dev/lvm-operater:latest
```
nbalacha, line 7: typo: operator

nbalacha (changes requested):

Dockerfile
```diff
@@ -13,15 +13,22 @@ RUN go mod download
...
 COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/vgmanager .
 USER 65532:65532
 
-ENTRYPOINT ["/manager"]
+# '/manager' is lvm-operator entrypoint
+# ENTRYPOINT ["/manager"]
+
+# '/vgmanager' is vgmanager entrypoint which is used in daemonset image
+# ENTRYPOINT ["/vgmanager"]
```
nbalacha, line 34: I think you can only have a single entrypoint per container.

nbalacha: > I think you can only have a single entrypoint per container image

leelavg: - we are invoking different binaries in `command` field of pod spec.

nbalacha: The entrypoint is the default command that is invoked so there should be only one. I think that will still allow us to use the vgmanager command in a different container explicitly.

leelavg: Yes, I thought explicit is better than implicit and commented out both entries, they are there just for ref

nbalacha: As per https://docs.docker.com/engine/reference/builder/ we will need an entrypoint. Please uncomment the one for the manager

config/manager/manager.env
```diff
@@ -4,4 +4,4 @@ CSI_REGISTRAR_IMAGE=k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
 CSI_PROVISIONER_IMAGE=k8s.gcr.io/sig-storage/csi-provisioner:v3.0.0
 CSI_LIVENESSPROBE_IMAGE=k8s.gcr.io/sig-storage/livenessprobe:v2.5.0
 CSI_RESIZER_IMAGE=k8s.gcr.io/sig-storage/csi-resizer:v1.3.0
-VGMANAGER_IMAGE=quay.io/ocs-dev/vgmanager:latest
+VGMANAGER_IMAGE=quay.io/ocs-dev/lvm-operator:latest
```
sp98, line 7: do we need `VGMANAGER_IMAGE` env if we are fetching the image from the container using `getRunningPodImage`?

leelavg: Yes, incase if you want to custom build only vgmanager and use that image.

sp98: custom build of vgmanager would mean building the operator image again (since vgmanager is part of operator build)

sp98: So if the primary use case is to be able to test custom vgmanager builds, then having separate images for both operator and vgmanager would be an ideal option, IMO.

controllers/vgmanager_daemonset.go
```diff
@@ -112,17 +117,30 @@ var (
...
 	volumes := []corev1.Volume{LVMDConfVol, DevHostDirVol, UDevHostDirVol, SysHostDirVol}
 	volumeMounts := []corev1.VolumeMount{LVMDConfVolMount, DevHostDirVolMount, UDevHostDirVolMount, SysHostDirVolMount}
 	privileged := true
 	var zero int64 = 0
+
+	// try to get running pod image, on failure fallback to default image
+	image, err := getRunningPodImage(r, ctx)
+	if err != nil {
+		r.Log.Error(err, "failed to get image from running operator")
+		image = VGManagerImage
```
sp98, line 132: IMO, we should either use `env` or use `getRunningPodImage`. We don't need both.

config/manager/manager.env
leelavg: - yes, you'll be commenting out vgmanager build and remove env in manager yaml. I can remove this but would like to have this atleast for a week or so and if we aren't seeing it's usage, I'll remove it.
- sounds good or want me to remove it now itself?

controllers/vgmanager_daemonset.go
sp98: and this would also mean we would never be able to test the custom vgmanager image in the `env`. Since it will always use the `lvm-operator` image!

controllers/vgmanager_daemonset.go
```diff
@@ -187,5 +205,33 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
...
-	return ds
+	return ds, nil
+}
+
+func getRunningPodImage(r *LVMClusterReconciler, ctx context.Context) (string, error) {
+
+	// 'POD_NAME' and 'WATCH_NAMESPACE' are set in env of lvm-operator when running as a container
+	podName := os.Getenv("POD_NAME")
+	if podName == "" {
+		return "", fmt.Errorf("failed to get image from running pod")
```
sp98, line 216:
```suggestion
		return "", fmt.Errorf("failed to get pod name env variable")
```

controllers/vgmanager_daemonset.go
```diff
@@ -187,5 +205,33 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
...
+	}
+
+	podNamespace := os.Getenv("WATCH_NAMESPACE")
+	if podNamespace == "" {
+		return "", fmt.Errorf("failed to get namespace from running pod")
```
sp98, line 221:
```suggestion
		return "", fmt.Errorf("failed to get namespace env variable")
```

controllers/vgmanager_daemonset.go
```diff
@@ -187,5 +205,33 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
...
+	}
+
+	pod := &corev1.Pod{}
+	if err := r.Get(ctx, types.NamespacedName{Name: podName, Namespace: podNamespace}, pod); err != nil {
+		return "", fmt.Errorf("failed to get pod spec of pod %s in namespace %s", podName, podNamespace)
```
sp98, line 226:
```suggestion
		return "", fmt.Errorf("failed to get pod %s in namespace %s", podName, podNamespace)
```

sp98 (changes requested):
nit

Makefile
```diff
@@ -98,22 +98,29 @@ test: manifests generate fmt vet envtest ## Run tests.
...
 	go build -o bin/manager main.go
 
-build-vgmanager: generate fmt vet ## Build manager binary.
+build-vgmanager: generate fmt vet ## Build vg manager binary.
 	go build -o bin/vgmanager cmd/vgmanager/main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
-docker-build: test ## Build docker image with the manager.
```
nbalacha, line 107: Can we leave this in? It is a good way to make sure everything passes.

controllers/vgmanager_daemonset.go
```diff
@@ -112,17 +117,36 @@ var (
...
+func newVGManagerDaemonset(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) (appsv1.DaemonSet, error) {
 	// aggregate nodeSelector and tolerations from all deviceClasses
 	nodeSelector, tolerations := extractNodeSelectorAndTolerations(lvmCluster)
 	volumes := []corev1.Volume{LVMDConfVol, DevHostDirVol, UDevHostDirVol, SysHostDirVol}
 	volumeMounts := []corev1.VolumeMount{LVMDConfVolMount, DevHostDirVolMount, UDevHostDirVolMount, SysHostDirVolMount}
 	privileged := true
 	var zero int64 = 0
+
+	// try to get vgmanager image from env and on absence get from running pod
+	// TODO: investigate why storing this env in a variable is failing tests
```
nbalacha, line 129: Is the test failure still a problem?

Makefile
leelavg: - As part of CI we are caching `go mod` and running `test` as standalone
- if docker-build also triggers `test` it's futile and only lengthens time

controllers/vgmanager_daemonset.go
leelavg: with current implementation it isn't, it happens when we store env values in a variable and use that here

controllers/defaults.go
```diff
@@ -100,6 +101,9 @@ var (
 
 	// default fstype for topolvm storage classes
 	TopolvmFilesystemType = "xfs"
+
+	// name of the lvm-operator container
+	ContainerName = "manager"
```
nbalacha, line 106: Please rename this to LVMOperatorContainerName.

controllers/vgmanager_daemonset.go
```diff
@@ -187,5 +211,33 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster, namespace string)
...
+
+func getRunningPodImage(r *LVMClusterReconciler, ctx context.Context) (string, error) {
+
+	// 'POD_NAME' and 'POD_NAMESPACE' are set in env of lvm-operator when running as a container
+	podName := os.Getenv("POD_NAME")
+	if podName == "" {
+		return "", fmt.Errorf("failed to get pod name env variable")
+	}
+
+	podNamespace := os.Getenv("POD_NAMESPACE")
```
nbalacha, line 225: This can be read directly from r.Namespace.

nbalacha (approved):

PR #39, opened 2021-12-22
fix: use default socket paths from topolvm project

leelavg:
- Make use of default path mentioned in topolvm project for CSI Socket
  and LVMd Socket files

Signed-off-by: Leela Venkaiah G <lgangava@redhat.com>

PR #46, opened 2021-12-24
fix: use default socket paths for CSI and LVMd

leelavg:
- In #39 values for socket paths are pulled from topolvm project
- However the directories for CSI and LVMd dock paths is same
- In #39 one of the dir is removed erroneously and that is fixed by this
  PR to use our own values not to have any clashes.

### DIFF
--- a/controllers/defaults.go
+++ b/controllers/defaults.go
@@ -18,8 +18,6 @@ package controllers
 
 import (
 	"os"
-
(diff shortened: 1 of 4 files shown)


sp98 (approved):

jmolmo (approved):

nbalacha (approved):
/lgtm

PR #53, opened 2021-12-30
feat: cleanup volume groups on CR deletion

leelavg:

### DIFF
--- a/config/rbac/vg_manager_role.yaml
+++ b/config/rbac/vg_manager_role.yaml
@@ -33,3 +33,19 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - topolvm.cybozu.com
+  resources:
+  - logicalvolumes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
--- a/pkg/vgmanager/cleanup.go
+++ b/pkg/vgmanager/cleanup.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vgmanager
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/red-hat-storage/lvm-operator/controllers"
+	topolvmv1 "github.com/topolvm/topolvm/api/v1"
+	lvmCMD "github.com/topolvm/topolvm/pkg/lvmd/cmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+// vgCleanupAll removes all volume groups existing in lvmd config file and after
+// removing every vg it also removes lvmd config file
+func vgCleanupAll(r *VGReconciler, ctx context.Context) error {
+
+	// get limited number of logical volume CRs to verify their presence
+	lvs := &topolvmv1.LogicalVolumeList{}
+	// TODO: perform cleanup based on individual volume groups after comparing
+	// lvmd config file with lvmcluster CR
+	if err := r.Client.List(ctx, lvs, &client.ListOptions{Limit: int64(1)}); err == nil {
+		// do not proceed with cleanup if there are logicalVolumes
+		if len(lvs.Items) > 0 {
+			return fmt.Errorf("found presence of logicalvolumes and aborting volume group cleanup")
+		}
+	} else {
+		r.Log.Error(err, "failed to list logicalvolumes CRs")
+		return err
+	}
+
+	// get all the volume groups from lvmd file
+	lvmdConfig := &lvmCMD.Config{
+		SocketName: controllers.DefaultLVMdSocket,
+	}
+
+	cfgBytes, err := os.ReadFile(controllers.LvmdConfigFile)
+	if os.IsNotExist(err) {
+		r.Log.Info("lvmd config file doesn't exist, no need of cleanup")
+		return nil
+	} else if err != nil {
+		r.Log.Error(err, "failed to read lvmd config file", "file path", controllers.LvmdConfigFile)
+		return err
+	} else {
+		err = yaml.Unmarshal(cfgBytes, &lvmdConfig)
+		if err != nil {
+			r.Log.Error(err, "failed to unmarshal yaml contents of lvmd config data")
+			return err
+		}
+	}
+
+	// remove volume group sequentially
+	var done int
+	isStarted := false
+	for done = 0; done < len(lvmdConfig.DeviceClasses); done++ {
+		deviceClass := lvmdConfig.DeviceClasses[done]
+		if err = RemoveVolumeGroup(deviceClass.Name); err != nil {
+			if errors.Is(err, ErrNotFound) {
+				// TODO: simplify the logic, multiple controllers are mutating same
+				// LVMCluster CR and as a result during the reconciliation cleanup is
+				// happening but failing to remove finalizer. So, if we try to remove
+				// a VG and it isn't found then that was probably deleted in previous
+				// reconile and so mark that the cleanup process was initiated
+				isStarted = true
+			}
+			r.Log.Error(err, "failed to remove volume group", "deviceClass", deviceClass.Name)
+			break
+		}
+		if !isStarted {
+			isStarted = true
+		}
+	}
+
+	if !isStarted && len(lvmdConfig.DeviceClasses) > 0 {
+		// not able to start cleanup, no need to overwrite config file, just return error
+		return fmt.Errorf("failed to initiate cleanup of Volume Groups")
+	} else if done == len(lvmdConfig.DeviceClasses)-1 || len(lvmdConfig.DeviceClasses) == 0 {
+		// every volume group is deleted and config file can be deleted
+		if err = os.Remove(controllers.LvmdConfigFile); err != nil {
+			return fmt.Errorf("failed to remove LVMd config file")
+		}
+	} else if done != 0 {
+		// some vgs are pending for deletion, overwrite config file
+		lvmdConfig.DeviceClasses = lvmdConfig.DeviceClasses[done:]
+		config, err := yaml.Marshal(lvmdConfig)
+		if err != nil {
+			r.Log.Error(err, "failed to marshal yaml contents of lvmd config data")
+			return err
+		}
+		if err = os.WriteFile(controllers.LvmdConfigFile, config, 0600); err != nil {
+			return fmt.Errorf("failed to overwrite LVMd config file")
+		}
+		pending := make([]string, 0, len(lvmdConfig.DeviceClasses))
+		for _, dc := range lvmdConfig.DeviceClasses {
+			pending = append(pending, dc.Name)
+		}
+		r.Log.Info("device classes are pending for deletion", "deviceClasses", pending)
+	}
+
+	return nil
+}
--- a/pkg/vgmanager/lvm.go
+++ b/pkg/vgmanager/lvm.go
@@ -28,17 +28,22 @@ import (
 	"github.com/red-hat-storage/lvm-operator/pkg/internal"
 )
 
+// TODO: Use json reporting format for LVM and remove parsing output line by line
+
 const (
 	// todo(rohan): make these paths configurable. I think they depend on the host OS
 	// this is why I copied the methods and didn't just import them from lvmd
 	lvmPath     = "/usr/sbin/lvm"
 	nsenterPath = "/usr/bin/nsenter"
+	vgremove    = "/usr/sbin/vgremove"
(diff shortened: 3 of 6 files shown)


pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// vgCleanup removes all volume groups existing in lvmd config file and on every vg
+// removal it also removed config file
+func vgCleanup(r *VGReconciler, ctx context.Context) error {
+
+	// get limited number of logical volume CRs to verify their presence
+	lvs := &topolvmv1.LogicalVolumeList{}
+	if err := r.Client.List(ctx, lvs, &crc.ListOptions{Limit: int64(1)}); err == nil {
+		// do not proceed with cleanup if there are logicalVolumes
```
sp98, line 38: Aren't we suppose to check if the LV CR is mapped with the VG somehow?  Right now we are aborting VG deletion if at least one LV CR is found.   This  won't be very useful in case of the cleaning up individual VGs (for example user removes a VG from the spec)

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -76,6 +82,37 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+			if err := r.Client.Update(ctx, lvmCluster); err != nil {
+				r.Log.Error(err, "failed to remove finalizer", "node", nodeName, "CR", lvmCluster.Name)
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// set node uid as finalizer on CR
+	if !controllerutil.ContainsFinalizer(lvmCluster, vgFinalizer) {
```
sp98, line 106: Can't we use `NodeName` itself as the finalizer ?

pkg/vgmanager/cleanup.go
leelavg: - that can be tweaked but I don't want to do that in this PR
- the flow will be more or less same however in controller itself there'll be some extra checks as well

pkg/vgmanager/vgmanager_controller.go
leelavg: - IMHO, UID is the standard way for these type of tasks.
- We can customize the finalizer to have prefix `node-uid/<uid>` for easy ref

pkg/vgmanager/cleanup.go
sp98: so the question would be even if the user is deleting the entire cluster, do we want to skip all VG deletion even if one LV CR is available? or we delete whatever VGs can be deleted in the cluster?

leelavg: Yes, until we have implementation of device selector and specific vg deletion, I'd like to abort any deletion.

When we implement the option 2, we can spend time on extra processing.

nbalacha: It would be better to have a vgCleanupAll function which calls a vgCleanup (vgname) for each vg.

config/rbac/vg_manager_role.yaml
```diff
@@ -41,3 +41,11 @@ rules:
   - securitycontextconstraints
   verbs:
   - use
+- apiGroups:
```
nbalacha, line 44: Was this generated by kubebuilder? If not please add that to the code.

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+		if len(lvs.Items) > 0 {
+			return fmt.Errorf("found presence of logicalvolumes and aborting volume group cleanup")
+		}
+	} else {
+		return err
+	}
+
+	// get all the volume groups from lvmd file
+	lvmdConfig := &lvmCMD.Config{
```
nbalacha, line 47: This should be a global var in the package or an argument to the function. Or an arg in the reconciler.

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+		SocketName: controllers.DefaultLVMdSocket,
+	}
+
+	cfgBytes, err := os.ReadFile(controllers.LvmdConfigFile)
+	if os.IsNotExist(err) {
+		r.Log.Info("lvmd config file doesn't exist, no need of cleanup")
+		return nil
+	} else if err != nil {
+		return err
```
nbalacha, line 56: Please log the error here.

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+	} else {
+		err = yaml.Unmarshal(cfgBytes, &lvmdConfig)
+		if err != nil {
+			return err
```
nbalacha, line 60: Please log the error.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -76,6 +82,37 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+			err := vgCleanup(r, ctx)
+			if err != nil {
+				r.Log.Error(err, "failed to perform VG Cleanup on node %s", nodeName)
+				return ctrl.Result{}, err
+			}
+
+			// remove finalizer after cleanup
+			controllerutil.RemoveFinalizer(lvmCluster, vgFinalizer)
+			if err := r.Client.Update(ctx, lvmCluster); err != nil {
+				r.Log.Error(err, "failed to remove finalizer", "node", nodeName, "CR", lvmCluster.Name)
```
nbalacha, line 98:
```suggestion
				r.Log.Error(err, "failed to remove vg finalizer", "node", nodeName, "LVMCluster", lvmCluster.Name, "finalizer", vgFinalizer)
```

pkg/vgmanager/lvm.go
```diff
@@ -124,11 +137,55 @@ func ListVolumeGroups() ([]*VolumeGroup, error) {
...
+		})
 	}
 	return groups, nil
 }
 
+// RemoveVolumeGroup removes volume group and physical volumes that are part of it
+func RemoveVolumeGroup(name string) error {
+
+	// find volume group which contains name and list of PVs belonging to VG
+	vg, err := FindVolumeGroup(name)
```
nbalacha, line 161: Unrelated to this pr but don't export this function if it is not being used outside the pkg.

nbalacha (changes requested):

pkg/vgmanager/cleanup.go
leelavg: the current function does the same, it'll remove volume group one by one based on vgName in lvmd config file

config/rbac/vg_manager_role.yaml
leelavg: - no, it's custom added.
- anything generated by kubebuilder is used by operator and these are for vgmanager

controllers/vgmanager_daemonset.go
```diff
@@ -153,6 +153,14 @@ func newVGManagerDaemonset(lvmCluster lvmv1alpha1.LVMCluster) appsv1.DaemonSet {
 						},
 					},
 				},
+				{
+					Name: "NODE_UID",
+					ValueFrom: &corev1.EnvVarSource{
+						FieldRef: &corev1.ObjectFieldSelector{
+							FieldPath: "metadata.uid",
```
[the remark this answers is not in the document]
leelavg: - querying for node uid based on node name set from env and using that

pkg/vgmanager/lvm.go
leelavg: - there are many functions like this, will take this in a separate PR

pkg/vgmanager/cleanup.go
leelavg: why should this be the case? my rationale is, if we fail to delete any VG then we need to read remaining content from config file in next reconcile and then this struct has to be re-declared again.

nbalacha: This struct should be the same one created by the VGmanager when creating the file. Don't regenerate it each time.

nbalacha: Actually scratch the last comment. What we should be doing is saving the socket and filename that was used when the file was created and using those in the struct.

pkg/vgmanager/lvm_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"os/exec"
+	"testing"
+)
+
+// TODO: unit test creation of Volume Groups and replace bash script
+
+var fileName = "vg-pool-"
+var vgName = "test-vg"
+
+func setup(name string) string { //nolint:golint,unused
```
sp98, line 14: is the test creating a new VG and deleting that? 
If yes, then do we really need to do that?

leelavg: 1. Currently this test will not work in CI as `nsenter` doesn't work and being skipped, need to refactor that to be testable
2. The intention is to create a volume group manually (eventually do that via vgmanager functions) and remove that via function being used in this PR, incase if deletion fails teardown will remove it quietly

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+		// not able to start cleanup, no need to overwrite config file, just return error
+		return fmt.Errorf("failed to initiate cleanup of Volume Groups")
+	} else if done == len(lvmdConfig.DeviceClasses)-1 || len(lvmdConfig.DeviceClasses) == 0 {
+		// every volume group is deleted and config file can be deleted
+		if err = os.Remove(controllers.LvmdConfigFile); err != nil {
+			return fmt.Errorf("failed to remove LVMd config file")
+		}
+	} else if done != 0 {
+		// some vgs are pending for deletion, overwrite config file
+		lvmdConfig.DeviceClasses = lvmdConfig.DeviceClasses[done:]
```
jmolmo, line 101: maybet it could be interesting to log what VGs are pending for deletion

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+
+	// remove volume group sequentially
+	var done int
+	isStarted := false
+	for done = 0; done < len(lvmdConfig.DeviceClasses); done++ {
+		deviceClass := lvmdConfig.DeviceClasses[done]
+		if err = RemoveVolumeGroup(deviceClass.Name); err != nil {
+			if errors.Is(err, ErrNotFound) {
```
jmolmo, line 76: It would be nice to log the error for other cases than "not found", the `FindVolumeGroup` function can fail executing a command and we will lost the information of what happened.

jmolmo, line 76: It would be nice to log the error for other cases than "not found", the FindVolumeGroup function can fail executing a command and we will lost the information of what happened.

jmolmo (changes requested):
It seems ok to me, just take a look to my comment about the error log when findvolume fails.

jmolmo (approved):

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,119 @@
...
+		if !isStarted {
+			isStarted = true
+		}
+	}
+
+	if !isStarted && len(lvmdConfig.DeviceClasses) > 0 {
+		// not able to start cleanup, no need to overwrite config file, just return error
+		return fmt.Errorf("failed to initiate cleanup of Volume Groups")
+	} else if done == len(lvmdConfig.DeviceClasses)-1 || len(lvmdConfig.DeviceClasses) == 0 {
+		// every volume group is deleted and config file can be deleted
```
sp98, line 96: When one of the vg deletion request fails, will `done` be still equal to `len(lvmdConfig.DeviceClasses)-1` and we will still remove the config file ?

pkg/vgmanager/cleanup.go
```diff
@@ -0,0 +1,119 @@
...
+			return err
+		}
+		if err = os.WriteFile(controllers.LvmdConfigFile, config, 0600); err != nil {
+			return fmt.Errorf("failed to overwrite LVMd config file")
+		}
+		pending := make([]string, 0, len(lvmdConfig.DeviceClasses))
+		for _, dc := range lvmdConfig.DeviceClasses {
+			pending = append(pending, dc.Name)
+		}
+		r.Log.Info("device classes are pending for deletion", "deviceClasses", pending)
```
sp98, line 115: Do we really need another list `pending` over here or can be just use `lvmdConfig.DeviceClasses` again?  
```suggestion
		r.Log.Info("device classes are pending for deletion", "deviceClasses", lvmdConfig.DeviceClasses)
```

sp98 (commented):
So we are always assuming that we would delete all the VGs when the user uninstalls lvm operator.  Can there be scenarios where user wants to delete the operator, but still want to use Persistent Volumes created by operator. That is, delete the operator but don't delete any data.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -76,6 +83,47 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+				r.Log.Error(err, "failed to remove vg finalizer", "node", nodeName, "LVMCluster", lvmCluster.Name, "finalizer", vgFinalizer)
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// set node uid as finalizer on CR
+	if !controllerutil.ContainsFinalizer(lvmCluster, vgFinalizer) {
+		controllerutil.AddFinalizer(lvmCluster, vgFinalizer)
```
sp98, line 118: If each vgManager daemonset pod is adding its own finalizer to the LVMCluster CR, then we might have issues in cleaning up of the finalizers. For example: if NodeSelector of the Vgmanager daemonset is updated, we might have stale finalizers in the LVMCluster CR finalizer.

leelavg:
- closing in favour of #73

PR #56, opened 2021-12-31
fix: create all resources in the operator namespace

nbalacha:
This fix creates all resources in the operator namespace instead of the LVMCluster namespace. This is required if, in future, the operator will be extended to watch all namespaces.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -53,7 +53,7 @@ spec:
             cpu: 100m
             memory: 20Mi
         env:
-        - name: WATCH_NAMESPACE
+        - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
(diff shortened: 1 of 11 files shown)


config/manager/manager.yaml
```diff
@@ -57,6 +57,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: POD_NAMESPACE
```
leelavg, line 60: - any scenario where `POD_NAMESPACE` will be different to `WATCH_NAMESPACE`, if not we can use the same i guess?

leelavg (commented):
- can we remove `operator_namespace` vars from `defaults.go` and `manager.env`?

config/manager/manager.yaml
nbalacha: I thought that this may be something we might want to support in the future but I agree that this will be the same for now. I will remove WATCH_NAMESPACE.

leelavg (approved):
/lgtm

PR #64, opened 2022-01-06
fix: removes VGMANAGER_IMAGE from the configmap

nbalacha:
Remove the VGMANAGER_IMAGE env variable from the manager
configmap. The default behaviour will be to use the same image
for both the controller and the vgmanager.


leelavg (approved):
/lgtm

PR #72, opened 2022-01-14
feat:  adds the LVMVolumeGroup CRD and processing logic

nbalacha:
The LVMCluster reconciliation will now create individual LVMVolumeGroup objects for each deviceClass. The VGManager will now reconcile the LVMVolumeGroup instead of the LVMCluster CR.


controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+	lvmvgName = "lvmvg-manager"
+)
+
+type lvmVG struct{}
+
+func (c lvmVG) getName() string {
+	return lvmvgName
+}
+
+func (c lvmVG) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
```
sp98, line 40: (Not related to this PR). Just realised the `context` should usually be  the first argument of the function.

cmd/vgmanager/main.go
```diff
@@ -75,8 +75,9 @@ func main() {
 	}
 
 	if err = (&vgmanager.VGReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Nodename: os.Getenv("NODE_NAME"),
```
jmolmo, line 80: Where is set this env. variable?

sp98, line 80: do we need `Nodename` in the `VGReconciler` object ? It s used only when we update the `status`. So we can directly call `os.Getenv("NODE_NAME")` over there.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, vgcr := range vgcrs {
+		err := r.Client.Get(ctx, types.NamespacedName{Name: vgcr.Name, Namespace: vgcr.Namespace}, existingLvmVg)
+
+		if err != nil {
+			// already deleted in previous reconcile
+			if errors.IsNotFound(err) {
+				r.Log.Info("LVMVG deleted", "Name", vgcr.Name)
+				return nil
+			}
+			r.Log.Error(err, "failed to retrieve LVMVG resource", "Name", vgcr.Name)
```
jmolmo, line 80: ":s/Name/name" in log lines

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	for _, vgcr := range vgcrs {
+		err := r.Client.Get(ctx, types.NamespacedName{Name: vgcr.Name, Namespace: vgcr.Namespace}, existingLvmVg)
+
+		if err != nil {
+			// already deleted in previous reconcile
+			if errors.IsNotFound(err) {
+				r.Log.Info("LVMVG deleted", "Name", vgcr.Name)
```
jmolmo, line 77: ":s/Name/name"

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+				return nil
+			}
+			r.Log.Error(err, "failed to retrieve LVMVG resource", "Name", vgcr.Name)
+			return err
+		}
+
+		// if not deleted, initiate deletion
+		if existingLvmVg.GetDeletionTimestamp().IsZero() {
+			if err = r.Client.Delete(ctx, existingLvmVg); err != nil {
+				r.Log.Error(err, "failed to delete LVMVG", "Name", existingLvmVg.Name)
```
jmolmo, line 87: ditto

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+				return err
+			} else {
+				r.Log.Info("initiated LVMVG deletion", "Name", existingLvmVg.Name)
```
jmolmo, line 90: ditto

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -46,7 +46,7 @@ const (
 func (r *VGReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.deviceAgeMap = newAgeMap(&wallTime{})
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&lvmv1alpha1.LVMCluster{}).
+		For(&lvmv1alpha1.LVMVolumeGroup{}).
```
sp98, line 49: Will creation of a new `LVMVolumeGroup` trigger controller reconcile for all the other existing LVMVolumeGroups?
If yes, then I think we should avoid that.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -71,18 +72,66 @@ func (r *VGReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+var reconcileAgain ctrl.Result = ctrl.Result{Requeue: true, RequeueAfter: reconcileInterval}
+
+//TODO: Refactor this function to move the ctrl result to a single place
+
 func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	lvmCluster := &lvmv1alpha1.LVMCluster{}
-	err := r.Client.Get(ctx, req.NamespacedName, lvmCluster)
+
+	var status string = "available"
+	var reason string = "Success"
```
jmolmo, line 84: ":/s/Success/success"

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -71,18 +72,66 @@ func (r *VGReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+	// filter out block devices
+	remainingValidDevices, delayedDevices, err := r.filterAvailableDevices(blockDevices)
+	if err != nil {
+		_ = err
+	}
+
+	var matchingDevices []internal.BlockDevice
+	_, matchingDevices, err = filterMatchingDevices(remainingValidDevices, volumeGroup)
+	if err != nil {
+		r.Log.Error(err, "could not filter matching Devices", "VGName", volumeGroup.Name)
```
jmolmo, line 124: ":S/Devices/devices"

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -111,64 +160,55 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
-			lvmdConfig.DeviceClasses = append(lvmdConfig.DeviceClasses, &lvmd.DeviceClass{
-				Name:        deviceClass.Name,
-				VolumeGroup: deviceClass.Name,
-				Default:     true,
-			})
+			status = "unavailable"
+			reason = "failed to create VG"
+		} else {
+			status = "degraded"
+			reason = "Failed to extend VG"
```
jmolmo, line 175: ":S/Failed/failed"

cmd/vgmanager/main.go
nbalacha: I don't see a problem with saving it. Similar to how we save the namespace elsewhere.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return nil
+}
+
+func (c lvmVG) updateStatus(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+	// intentionally empty
+	return nil
+}
+
+func (c lvmVG) getLvmVolumeGroups(r *LVMClusterReconciler, instance *lvmv1alpha1.LVMCluster) ([]*lvmv1alpha1.LVMVolumeGroup, error) {
```
sp98, line 106: We can only pass the `namespace` instead of entire `LVMClusterReconciler`, if we only plan to use the `namespace`

jmolmo (approved):
Just minor things related with the use of Caps letters in log messages

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+
+	lvmVolumeGroups := []*lvmv1alpha1.LVMVolumeGroup{}
+
+	deviceClasses := instance.Spec.DeviceClasses
+	for _, deviceClass := range deviceClasses {
+		lvmVolumeGroup := &lvmv1alpha1.LVMVolumeGroup{
```
sp98, line 112: how do we decide which `deviceClass` is `default`?

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+			Spec: lvmv1alpha1.LVMVolumeGroupSpec{
+				NodeSelector:   deviceClass.NodeSelector,
+				DeviceSelector: deviceClass.DeviceSelector,
+			},
+		}
+		lvmVolumeGroups = append(lvmVolumeGroups, lvmVolumeGroup)
+	}
+	return lvmVolumeGroups, nil
+
```
sp98, line 125: blank line.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	cutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	lvmvgName = "lvmvg-manager"
```
sp98, line 31:
```suggestion
	lvmVGName = "lvmvg-manager"
```

sp98 (changes requested):

cmd/vgmanager/main.go
sp98: [here](https://github.com/red-hat-storage/lvm-operator/blob/b76aeb2f19aa0aceacf955c56eb70d7c661f468e/controllers/vgmanager_daemonset.go#L157).

sp98: ok.  name should be `NodeName`.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	for _, volumeGroup := range lvmVolumeGroups {
+		result, err := cutil.CreateOrUpdate(ctx, r.Client, volumeGroup, func() error {
+			// no need to mutate any field
+			return nil
+		})
+
+		if err != nil {
+			r.Log.Error(err, "LvmVolumeGroup reconcile failure", "name", volumeGroup.Name)
```
sp98, line 54: For consistency

```suggestion
			r.Log.Error(err, "failed to reconcile LvmVolumeGroup", "name", volumeGroup.Name)
```

controllers/lvm_volumegroup.go
nbalacha: The Name is the fieldname. That will be capitalized

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+
+	existingLvmVg := &lvmv1alpha1.LVMVolumeGroup{}
+	vgcrs, err := c.getLvmVolumeGroups(r, lvmCluster)
+	if err != nil {
+		return err
+	}
+
+	for _, vgcr := range vgcrs {
+		err := r.Client.Get(ctx, types.NamespacedName{Name: vgcr.Name, Namespace: vgcr.Namespace}, existingLvmVg)
+
```
sp98, line 73: blank line.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			// already deleted in previous reconcile
+			if errors.IsNotFound(err) {
+				r.Log.Info("LVMVG deleted", "Name", vgcr.Name)
+				return nil
```
sp98, line 78: shouldn't we `continue` here rather than returning nil ?

sp98 (changes requested):

controllers/lvm_volumegroup.go
sp98: Its not consistent through out.  Both `name` and `Name` are used in the logs.

pkg/vgmanager/vgmanager_controller.go
nbalacha: Status strings are usually capitalized.

controllers/lvm_volumegroup.go
sp98, line 106: `getLvmVolumeGroups` does not return any error. Should remove `error` from the return types.

controllers/lvm_volumegroup.go
sp98, line 77: logs should be more consistent.  `ensureDeleted` uses `LVMVG` in the log messages while `ensureCreated` uses `LvmVolumeGroup`

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -243,3 +284,32 @@ func ToleratesTaints(tolerations []corev1.Toleration, taints []corev1.Taint) boo
...
+
+	found := false
+
+	nodeStatuses := instance.Status.NodeStatus
+	for _, nodeStatus := range nodeStatuses {
+		if nodeStatus.Node == r.Nodename {
+			found = true
+			nodeStatus.Status = status
+			nodeStatus.Reason = reason
+		}
```
sp98, line 298: `break` for loop once `NodeName` is found.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,126 @@
...
+			r.Log.Error(err, "failed to retrieve LVMVG resource", "Name", vgcr.Name)
+			return err
+		}
+
+		// if not deleted, initiate deletion
+		if existingLvmVg.GetDeletionTimestamp().IsZero() {
+			if err = r.Client.Delete(ctx, existingLvmVg); err != nil {
+				r.Log.Error(err, "failed to delete LVMVG", "Name", existingLvmVg.Name)
+				return err
+			} else {
```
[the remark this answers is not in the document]
nbalacha: Done

controllers/lvm_volumegroup.go
nbalacha: done

controllers/lvm_volumegroup.go
nbalacha: Ack. Changing it to "name".

controllers/lvm_volumegroup.go
nbalacha: Done.

nbalacha: Done

controllers/lvm_volumegroup.go
nbalacha: Done.

controllers/lvm_volumegroup.go
nbalacha: I'd like to address that in a separate PR.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -243,3 +284,32 @@ func ToleratesTaints(tolerations []corev1.Toleration, taints []corev1.Taint) boo
...
+	}
+	if !found {
```
sp98, line 300: I don't think it will update the `Status` when the `NodeName` was `found` but it has different `Status` or `Reason`.  So we should add status if NodeName not found and update status if NodeName is found but its `Status` or `Reason` are different.

sp98 (changes requested):

controllers/lvm_volumegroup.go
nbalacha: done.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -111,64 +160,55 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
-			continue
-		}
-		_, found := deviceClassMap[deviceClass.Name]
 		if !found {
-			lvmdConfig.DeviceClasses = append(lvmdConfig.DeviceClasses, &lvmd.DeviceClass{
-				Name:        deviceClass.Name,
-				VolumeGroup: deviceClass.Name,
-				Default:     true,
-			})
+			status = "unavailable"
```
sp98, line 171: `Status` should be used as constants (enums) so that its easier to know what all statuses are available.

controllers/lvm_volumegroup.go
nbalacha: done.

controllers/lvm_volumegroup.go
nbalacha: Done

controllers/lvm_volumegroup.go
nbalacha: done.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -243,3 +284,32 @@ func ToleratesTaints(tolerations []corev1.Toleration, taints []corev1.Taint) boo
 	}
 	return true
 }
+
+func (r *VGReconciler) updateStatus(ctx context.Context, status string, reason string, instance *lvmv1alpha1.LVMVolumeGroup) error {
+
+	found := false
```
sp98, line 290: We still have a situation where multiple controllers will be updating the same `LVMVolumeGroup` status, right?  
If that's the case, should we first get the latest `LVMVolumeGroup` instance before verifying the `NodeStatus`?

cmd/vgmanager/main.go
nbalacha: Done

pkg/vgmanager/vgmanager_controller.go
nbalacha: Done

pkg/vgmanager/vgmanager_controller.go
nbalacha: Done

pkg/vgmanager/vgmanager_controller.go
nbalacha: Didn't understand this.

pkg/vgmanager/vgmanager_controller.go
sp98: I think this maybe be enough, to ensure that we are using the updated instance of `LVMVolumeGroup`.

pkg/vgmanager/vgmanager_controller.go
sp98: We are only updating the `instance.Status.NodeStatus` in line 307, inside the ` if !found` block.   What if `found` is `true` but `Status` and/or `Reason` is different. In this case, the `Status` and `Reason` will never be updated.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -111,64 +160,55 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
-		}
-		_, found := deviceClassMap[deviceClass.Name]
 		if !found {
-			lvmdConfig.DeviceClasses = append(lvmdConfig.DeviceClasses, &lvmd.DeviceClass{
-				Name:        deviceClass.Name,
-				VolumeGroup: deviceClass.Name,
-				Default:     true,
-			})
+			status = "unavailable"
+			reason = "failed to create VG"
```
sp98, line 172: `reason` for `unavailable` starts with small `f`, while that of `degraded` starts with capital `F`

pkg/vgmanager/vgmanager_controller.go
nbalacha: The status is being updated in the original array if the node is found. append should create a new slice which is why it is being assigned again.

pkg/vgmanager/vgmanager_controller.go
nbalacha: That can be taken up in a different PR.

pkg/vgmanager/vgmanager_controller.go
nbalacha: If it is not the latest version, I expect the update to fail. I will confirm after testing on a multi node cluster.

pkg/vgmanager/vgmanager_controller.go
sp98: got it.

sp98 (approved):

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (c lvmVG) getName() string {
+	return lvmVGName
+}
+
+func (c lvmVG) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCluster *lvmv1alpha1.LVMCluster) error {
+
+	lvmVolumeGroups := c.getLvmVolumeGroups(r, lvmCluster)
+
+	for _, volumeGroup := range lvmVolumeGroups {
```
travisn, line 43: Creating the LVMVolumeGroup CRs from the LVMCluster CR is still a design concern. The admin should create the LVMVolumeGroup CRs directly. Not a blocker for feature complete, but we really need to discuss this.

nbalacha: I disagree. The admin should create the LVMCluster CR and the controller should create the LVMVolumeGroup CRs. The intention is to have a single CR for the controller to create the internal pieces from.

sp98:
Q:  When we add the functionality to read any new device add events on the nodes, who will we tell all the LVMVolumeGroup CRs  associated with the node to reconcile?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -71,18 +75,60 @@ func (r *VGReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+
+	volumeGroup := &lvmv1alpha1.LVMVolumeGroup{}
+	err := r.Client.Get(ctx, req.NamespacedName, volumeGroup)
 	if err != nil {
+		r.Log.Error(err, "failed to get LVMVolumeGroup", "VGName", req.Name)
 		return ctrl.Result{}, err
 	}
 
-	r.Log.Info("listing block devices")
+	//TODO: actually check the node against the nodeSelector.
```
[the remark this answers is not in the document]
nbalacha: The CR is the object stored in etcd and does not run on a node. The vgmanager pod is expected to check the node it is running on against the nodeselector specified in the LVMVolumeGroup CR to see if it expected to act on it. This is currently not being done.

nbalacha:
> Q: When we add the functionality to read any new device add events on the nodes, how will we tell all the LVMVolumeGroup CRs associated with the node to reconcile?

The VG manager will need to reread the CRs once it receives the event. No changes will be made to the CR to trigger a reconcile.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -71,18 +75,60 @@ func (r *VGReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+	var matchingDevices []internal.BlockDevice
+	_, matchingDevices, err = filterMatchingDevices(remainingValidDevices, volumeGroup)
+	if err != nil {
+		r.Log.Error(err, "could not filter matching devices", "VGName", volumeGroup.Name)
+		return reconcileAgain, err
+	}
+
+	status := &lvmv1alpha1.VGStatus{
+		Name:   req.Name,
+		Status: lvmv1alpha1.VGStatusReady,
```
leelavg, line 130: - IMHO initial vgstatus should be unknown and we try to set it correctly by the end of reconcile 🤔

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -111,64 +157,72 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
 
 	// apply and save lvmconfig
 	// pass config to configChannel only if config has changed
 	if !cmp.Equal(existingLvmdConfig, lvmdConfig) {
 		r.Log.Info("updating lvmd config")
 		out, err := yaml.Marshal(lvmdConfig)
-		if err != nil {
-			return ctrl.Result{}, err
+		if err == nil {
+			err = os.WriteFile(controllers.LvmdConfigFile, out, 0600)
```
leelavg, line 211: why not move this line to 218 and remove this conditional?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -243,3 +298,71 @@ func ToleratesTaints(tolerations []corev1.Toleration, taints []corev1.Taint) boo
...
+			Name:      r.NodeName,
+			Namespace: r.Namespace,
+		},
+	}
+
+	result, err := ctrl.CreateOrUpdate(ctx, r.Client, nodeStatus, func() error {
+		if nodeStatus.CreationTimestamp.IsZero() {
+			vgNodeStatus.DeepCopyInto(nodeStatus)
+			return nil
+		}
```
leelavg, line 341: - i think `setStatus` already handles this condition by creating a new status when one is not found?

pkg/vgmanager/vgmanager_controller.go
nbalacha: Can this be taken up in a separate PR?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -111,64 +157,72 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
-		if !(selectsNode && ToleratesTaints(deviceClass.Tolerations, node.Spec.Taints)) {
-			continue
+	if len(matchingDevices) == 0 {
+		r.Log.Info("no matching devices", "VGName", volumeGroup.Name)
+		if len(delayedDevices) > 0 {
+			return reconcileAgain, nil
+		} else {
+			if found {
+				// Update the status again just to be safe.
+				if statuserr := r.updateStatus(ctx, status, volumeGroup); statuserr != nil {
```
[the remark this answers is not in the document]
nbalacha: Yes, because the lvmd.yaml file is only updated once the VG is created. This block is mainly to handle cases when the status might not have been updated correctly. That function can also be modified to not update if there is no change for this entry.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -71,18 +75,60 @@ func (r *VGReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to list block devices: %v", err)
+		return reconcileAgain, fmt.Errorf("failed to list block devices: %v", err)
+	}
+
+	// filter out block devices
+	remainingValidDevices, delayedDevices, err := r.filterAvailableDevices(blockDevices)
+	if err != nil {
+		_ = err
+	}
```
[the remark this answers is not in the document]
nbalacha: This was the earlier existing code which I have not changed.

controllers/lvm_volumegroup.go
```diff
@@ -0,0 +1,114 @@
...
+		result, err := cutil.CreateOrUpdate(ctx, r.Client, volumeGroup, func() error {
+			// no need to mutate any field
```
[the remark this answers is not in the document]
nbalacha: Correct. For now this is fine. The comment can be removed when that is actually done.

pkg/vgmanager/vgmanager_controller.go
nbalacha: The checks work like this:
if unmarshall fails,-> return error
If unmarshall succeeds -> attempt to write
if write fails -> return error

Notice that there is no else check - so the err check will work for both scenarios.

pkg/vgmanager/vgmanager_controller.go
nbalacha: This is working on the existing CR in the system so we need to update that object.

leelavg (approved):
/lgtm

controllers/lvm_volumegroup.go
travisn: I still haven't seen a design backing this decision, so my position remains to disagree with it so we can follow standard K8s patterns.

PR #75, opened 2022-01-17
feat:  move tolerations into LVMCluster.Spec

nbalacha:
This PR moves the tolerations field out of the deviceClass and into the main LVMCluster.Spec.
It also updates the tolerations in topolvm-node daemonset to match that of the vgmanager daemonset.


controllers/utils.go
```diff
@@ -25,11 +25,13 @@ import (
 // extractNodeSelectorAndTolerations combines and extracts scheduling parameters from the multiple deviceClass entries in an lvmCluster
 func extractNodeSelectorAndTolerations(lvmCluster *lvmv1alpha1.LVMCluster) (*corev1.NodeSelector, []corev1.Toleration) {
 	var nodeSelector *corev1.NodeSelector
-	var tolerations []corev1.Toleration
+
+	tolerations := lvmCluster.Spec.Tolerations
```
sp98, line 29: how about just returning `lvmCluster.Spec.Tolerations` from this method and not saving it in a variable first. The variable is never used.   

Or just have `extractNodeSelector` method. We can directly use `lvmCluster.Spec.Tolerations` if possible.

jmolmo (approved):

sp98 (approved):

controllers/utils.go
nbalacha: While that is a valid comment,I would like to address that in a separate PR

PR #80, opened 2022-01-18
fix: sets bundle namespace to openshift-storage

nbalacha:
This sets the bundle manifests namespace to openshift-storage by
default.


leelavg (approved):
/lgtm

PR #89, opened 2022-01-27
fix: sets env vars in the manager deployment

nbalacha:
Env vars for the manager deployment were earlier set using a configmap.
That does not work well with the downstream build mirroring pipeline.
This fix sets the env vars directly in the manager deployment spec.


sp98 (approved):

leelavg (approved):
/lgtm

PR #94, opened 2022-01-28
fix: use operator image for init containers

leelavg:
- move function to get running pod image into common controller utils
- use operator image from running pod for init containers in topolvm
  controller and topolvm node

### DIFF
--- a/controllers/vgmanager_daemonset.go
+++ b/controllers/vgmanager_daemonset.go
@@ -18,14 +18,12 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"os"
 
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


controllers/utils.go
```diff
@@ -56,3 +61,26 @@ func setDaemonsetNodeSelector(nodeSelector *corev1.NodeSelector, ds *appsv1.Daem
 		ds.Spec.Template.Spec.Affinity = nil
 	}
 }
+
+func getRunningPodImage(r *LVMClusterReconciler, ctx context.Context) (string, error) {
+
```
nbalacha, line 66: I would recommend saving the image in the reconciler struct instead of doing this each time.

controllers/vgmanager_daemonset.go
```diff
@@ -128,13 +126,8 @@ func newVGManagerDaemonset(r *LVMClusterReconciler, ctx context.Context, lvmClus
 	// try to get vgmanager image from env and on absence get from running pod
 	// TODO: investigate why storing this env in a variable is failing tests
 	image := os.Getenv("VGMANAGER_IMAGE")
```
nbalacha, line 128: I think this can be removed and image can be set to r.ImageName

leelavg: can i take this up during removal of standalone vg_manager?

nbalacha (approved):

controllers/utils.go
leelavg: done.

Issue #29, opened 2021-12-20
Use PSPs/Pod security standards to restrict unwanted resource/namespace access to host


leelavg:
- With SCC we can only target namespace level isolation and SELinux, Syscalls filteration
- Add PSP to further restrict this access, for example restricting access to a single directory when using hostPath

Madhu-1:
@leelavg PSP is deprecated in kubernetes 1.21 we should not use it anyone and look for the alternative like pod security https://kubernetes.io/docs/concepts/security/pod-security-standards ?

leelavg:
> PSP is deprecated in kubernetes 1.21 we should not use it anyon
- ack, we are aware of this and so didn't enforce PSPs, as there was no alternative mentioned in PSP blog post (or atleast I missed to notice) this issue is a placeholder wrt security

> alternative like pod security https://kubernetes.io/docs/concepts/security/pod-security-standards ?
- thanks, will look into.

suleymanakbas91:
We will track this at https://issues.redhat.com/browse/OCPVE-344. Thanks @leelavg!

PR #101, opened 2022-02-02
docs: add dev guide explaining lvmo components [skip ci]

leelavg:

### DIFF
--- a/doc/dev-guide/lvmo-units.md
+++ b/doc/dev-guide/lvmo-units.md
@@ -1,0 +1,23 @@
+## LVM Volume Groups
+
+- *lvmVG* reconcile units deploys and manages LVMVolumeGroup CRs
+- The LVMVG resource manager creates individual LVMVolumeGroup CRs for each
+  deviceClass in the LVMCluster CR. The vgmanager watches the LVMVolumeGroup
+  and creates the required volume groups on the individual nodes based on the
+  specified deviceSelector and nodeSelector.
+- The corresponding CRs forms the basis of `vgManager` unit to create volume
+  groups and create lvmd config file
+
+## Openshift SCCs
+
+- When the operator is deployed on an Openshift cluster all the required
+  SCCs are created by `openshiftSccs` reconcile unit
+- The `vg-manager`, `topolvm-node` and `lvmd` containers need elevated
+  permissions to access host LVM commands using `nsenter`.
+
+## Storage Classes
+
+- *topolvmStorageClass* resource units creates and manages all the storage
+  classes corresponding to the deviceClasses in the LVMCluster
+- Storage Class name is generated with a prefix "topolvm-" added to name of the
+  device class in the LVMCluster CR
--- a/doc/dev-guide/reconciler.md
+++ b/doc/dev-guide/reconciler.md
@@ -1,0 +1,59 @@
+# Operator design
+
+## Controllers and their managed resources
+
+### lvmcluster-controller
+
+- On receiving a valid LVMCluster CR, lvmcluster-controller reconciles the
+  following resource units for setting up [Topolvm](topolvm-repo) CSI and all
+  the supporting resources to use storage local to the node via Logical Volume
+  Manager (lvm)
+- *csiDriver*: Reconciles topolvm CSI Driver
+- *topolvmController*: Reconciles topolvm controller plugin
+- *lvmVG*: Reconciles volume groups from LVMCluster CR
+- *openshiftSccs*: Manages SCCs when the operator is run in Openshift
+  environment
+- *topolvmNode*: Reconciles topolvm nodeplugin along with lvmd
+- *vgManager*: Responsible for creation of Volume Groups
+- *topolvmStorageClass*: Manages storage class life cycle based on
+  devicesClasses in LVMCluster CR
+- The LVMO creates an LVMVolumeGroup CR for each deviceClass in the
+  LVMCluster CR. The LVMVolumeGroups are reconciled by the vgmanager controllers.
+- In addition to managing above resource units, lvmcluster-controller collates
+  the status of deviceClasses across nodes from LVMVolumeGroupNodeStatus and
+  updates status of LVMCluster CR
+- `resourceManager` interface is defined in
+  [lvmcluster_controller.go][contorller]
+- Depending on the resource unit some of the methods can be no-op
+
+Note:
+- Above names refers to the struct which satisfies `resourceManager` interface
+- Please refer to topolvm [design][topolvm-design] doc to know more about Topolvm
+  CSI
+- Any new resource units should also implement `resourceManager` interface
+
+### Lifecycle of Custom Resources
+
+- [LVMCluster CR][lvmcluster] represents the volume groups that should be
+  created and managed across nodes with custom node selector, toleration and
+  device selectors
+- Should be created and edited by user in operator installed namespace
+- Only a single CR instance with a single volume group containing all available
+  disks across schedulable nodes is supported and implementation respecting
+  tolerations, device and node selector fields is coming soon
+- All fields in `status` are updated based on the status of volume groups
+  creation across nodes
+
+Note:
+- Device Class and Volume Group can be read interchangeably
+- Multiple CRs exist to separate concerns of which component deployed by LVMO
+  updates which CR there by reducing multiple reconcile loops and colliding
+  requests/updates to Kubernetes API Server
+- Feel free to raise a github [issue][issue] for open discussions about API
+  changes if required
+
+[topolvm-repo]: https://github.com/topolvm/topolvm
+[topolvm-design]: https://github.com/topolvm/topolvm/blob/main/docs/design.md
+[controller]: ../../controllers/lvmcluster_controller.go
+[lvmcluster]: ../../api/v1alpha1/lvmcluster_types.go
+[issue]: https://github.com/red-hat-storage/lvm-operator/issues
--- a/doc/dev-guide/topolvm-csi.md
+++ b/doc/dev-guide/topolvm-csi.md
@@ -1,0 +1,34 @@
+# Topolvm CSI
+
+- LVM Operator deploys the Topolvm CSI which provides dynamic provisioning of
+  local storage.
+- Please refer to topolvm [docs][topolvm-docs] for more details on topolvm
+
+## CSI Driver
+
+- *csiDriver* reconcile unit deploys Topolvm CSIDriver resource
+
+## Topolvm Controller
+
+- *topolvmController* reconcile unit deploys a single Topolvm Controller plugin
+  deployment and manages any updates to the deployment
+- Topolvm scheduler is not used for pod scheduling. The CSI StorageCapacity
+  tracking feature by the scheduler
+- An init container generates openssl certs to be used in topolvm controller
+  which will be soon replaced with cert-manager
+
+## Topolvm Node and LVMd
+
+- *topolvmNode* reconcile unit deploys and manages topolvm node plugin and lvmd
+  daemonset and scales it based on node selector specified in the devicesClasses
+  in LVMCluster
+- An init container polls for the availability of lvmd config file before
+  starting the lvmd and topolvm-node containers
+
+## Deletion
+
+- All above resources will be removed by their respective reconcile units when
+  LVM Cluster CR governing then is deleted
+
+
+[topolvm-docs]: https://github.com/topolvm/topolvm/tree/main/docs
--- a/doc/dev-guide/index.md
+++ b/doc/dev-guide/index.md
@@ -1,0 +1,13 @@
+# Contents
+
+The dev-guide describes various components managed by LVM Operator (LVMO)
+
+1. [Reconcile][reconciler]
+2. [VG Manager][vg-manager]
+3. [CSI Units][topolvm-csi]
+4. [LVMO Units][lvmo-units]
+
+[reconciler]: ./reconciler.md
+[vg-manager]: ./vg-manager.md
+[topolvm-csi]: ./topolvm-csi.md
+[lvmo-units]: ./lvmo-units.md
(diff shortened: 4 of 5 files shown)


doc/dev-guide/reconciler.md
```diff
@@ -0,0 +1,59 @@
...
+### lvmcluster-controller
+
+- On receiving a valid LVMCluster CR, lvmcluster-controller reconciles the
+  following resource units for setting up [Topolvm](topolvm-repo) CSI and all
+  the supporting resources to use storage local to the node via Logical Volume
+  Manager (lvm)
+- *csiDriver*: Reconciles topolvm CSI Driver
+- *topolvmController*: Reconciles topolvm controller plugin
+- *lvmVG*: Reconciles volume groups from LVMCluster CR
+- *openshiftSccs*: Manages SCCs when the operator is run in Openshift
```
travisn, line 14: The operator creates the SCCs? Security best practices is that the SCCs should be defined statically and not managed by an operator. For example, Rook defines its SCC in its upstream manifest [here](https://github.com/rook/rook/blob/master/deploy/examples/operator-openshift.yaml#L9). However, downstream doesn't have a way to define the SCCs in OLM, so the workaround was to have the operator do it. But upstream really should define the SCC in a standalone manifest so the operator doesn't need high privileges to create SCCs.

doc/dev-guide/lvmo-units.md
```diff
@@ -0,0 +1,23 @@
...
+  and creates the required volume groups on the individual nodes based on the
+  specified deviceSelector and nodeSelector.
+- The corresponding CRs forms the basis of `vgManager` unit to create volume
+  groups and create lvmd config file
+
+## Openshift SCCs
+
+- When the operator is deployed on an Openshift cluster all the required
+  SCCs are created by `openshiftSccs` reconcile unit
+- The `vg-manager`, `topolvm-node` and `lvmd` containers need elevated
```
travisn, line 15: Where is the RBAC defined? The SCCs don't define all the privileges.

doc/dev-guide/lvmo-units.md
```diff
@@ -0,0 +1,23 @@
...
+  permissions to access host LVM commands using `nsenter`.
+
+## Storage Classes
+
+- *topolvmStorageClass* resource units creates and manages all the storage
+  classes corresponding to the deviceClasses in the LVMCluster
+- Storage Class name is generated with a prefix "topolvm-" added to name of the
```
travisn, line 22: Can a storage class name be specified without the topolvm- prefix being added? Admins generally like to control the names of their storage classes.

doc/dev-guide/lvmo-units.md
```diff
@@ -0,0 +1,23 @@
+## LVM Volume Groups
+
+- *lvmVG* reconcile units deploys and manages LVMVolumeGroup CRs
+- The LVMVG resource manager creates individual LVMVolumeGroup CRs for each
+  deviceClass in the LVMCluster CR. The vgmanager watches the LVMVolumeGroup
+  and creates the required volume groups on the individual nodes based on the
+  specified deviceSelector and nodeSelector.
+- The corresponding CRs forms the basis of `vgManager` unit to create volume
+  groups and create lvmd config file
```
travisn, line 9: How is the lvmd config file generated on each node? Does the vgManager have a hostPath mounted? Could you add more details about that?

doc/dev-guide/reconciler.md
```diff
@@ -0,0 +1,59 @@
...
+  environment
+- *topolvmNode*: Reconciles topolvm nodeplugin along with lvmd
+- *vgManager*: Responsible for creation of Volume Groups
+- *topolvmStorageClass*: Manages storage class life cycle based on
+  devicesClasses in LVMCluster CR
+- The LVMO creates an LVMVolumeGroup CR for each deviceClass in the
```
travisn, line 20: This is the point I still don't understand... Why doesn't the admin create the LVMVolumeGroup CR directly? I don't see any description of what we're solving by embedding them in the cluster CR.

travisn (commented):
Great to see the design writeup, thanks!

doc/dev-guide/reconciler.md
leelavg: - i may be wrong here, does rook being deployed standalone in downstream, isn't that setup by meta operators which also takes care of SCCs?
- if indeed rook upstream isn't setting SCCs from code, it does have to deploy necessary RBAC required to access SCC api irrespective of upstream/downstream, isn't it?

doc/dev-guide/lvmo-units.md
leelavg: - yes, RBAC is either part of CSV or `make deploy`
- However, if you are referring to any security rbac, we are yet to tighten it as mentioned in #29

doc/dev-guide/lvmo-units.md
leelavg: - if admin doesn't want to use a pre-created storage class they are free to create a new storage class and not use pre-created SC at all
- we aren't setting any SC as default but just creating SC's corresponding to device classes

doc/dev-guide/lvmo-units.md
leelavg: - doc did explain about hostPath in the [initial](https://github.com/red-hat-storage/lvm-operator/pull/87/files#diff-a99023791c3c5bc7aa8c1977340936a533ee8c71aad9424dca56f093221ff57cR19-R25) version of this PR
- maybe I misunderstood that comment and removed that info from vg-manager doc

doc/dev-guide/reconciler.md
leelavg: - same as https://github.com/red-hat-storage/lvm-operator/pull/100#discussion_r798198723
- it's mostly about having to update status on the CR by multiple controllers which'll result into parallel CR update requests

doc/dev-guide/lvmo-units.md
travisn: I was primarily just referring to the RBAC so the operator and other controllers can have access to the K8s resources.

doc/dev-guide/lvmo-units.md
travisn: What about adding a property on the device class something like `storageClassName` that allows them to override the generated storage class name? It's just a suggestion for a future PR.

doc/dev-guide/reconciler.md
travisn: > * i may be wrong here, does rook being deployed standalone in downstream, isn't that setup by meta operators which also takes care of SCCs?

Downstream Rook is only deployed as part of ODF and the OCS operator creates the necessary SCC since Rook does not create its own. 

> * if indeed rook upstream isn't setting SCCs from code, it does have to deploy necessary RBAC required to access SCC api irrespective of upstream/downstream, isn't it?

The Rook operator does not ever create SCCs or RBAC. It's all defined in its helm chart (for helm users), or the [common.yaml](https://github.com/rook/rook/blob/master/deploy/examples/common.yaml) for other upstream users. So the admin has full control to grant the operator only exactly the permissions needed.

nbalacha: How should this be done for downstream outside of the operator? Bundles do not allow us to include SCCs so we decided to do this in the operator in order to make deployment easier.

doc/dev-guide/lvmo-units.md
nbalacha: The RBACs are part of the bundle (OCP) and manifests (k8s) in the config directory.

doc/dev-guide/lvmo-units.md
nbalacha: The intention here is to provide a 1-click install so there is a working storage setup the admin can use from day 1, similar to how OCS does this.  They can create their own sc later if they choose.

doc/dev-guide/reconciler.md
nbalacha: We want the operator to have the ability to generate the LVMVolumeGroup CRs based on the LVMCluster CR. This allows us better control and validation as discussed earlier.

doc/dev-guide/lvmo-units.md
travisn: Upstream users appreciate customizing from the first install. That's the beauty of CRs is that it allows for the customization from the start. Admins don't generally want to have extra/unused/uncustomizable storage classes. 1-click install is great for the common case, but requiring the 1-click install doesn't work well for the advanced users.

doc/dev-guide/reconciler.md
travisn: Downstream does require the operator to do it. I'm just suggesting that for upstream there should be an option for the operator not to do it, so the SCC can be created separately and avoid giving the operator the extra privileges to create SCCs.

doc/dev-guide/reconciler.md
travisn: > it's mostly about having to update status on the CR by multiple controllers which'll result into parallel CR update requests

The problem of updating the status seems the same either way. I would agree that the operator should own updating the status on the LVMVolumeGroup CRs, but that doesn't mean the admin can't create those CRs.

> We want the operator to have the ability to generate the LVMVolumeGroup CRs based on the LVMCluster CR. This allows us better control and validation as discussed earlier.

What control and validation is needed? These are the details I'm missing. It seems any meaningful validation must happen on nodes by the vgmanager, not in the operator. The operator doesn't have insight to storage actually available.

nbalacha: It could be something as simple as making sure there are not 2 entries trying to use up all the disks on all the nodes. It allows the operator to perform a first level validation before the vgmanager controller starts to create the VGs on the nodes. There could be other validations that we may perform in the future as the operator grows.

doc/dev-guide/lvmo-units.md
nbalacha: A good point and can be taken up in later improvements. Out first priority was to have a working operator and one of the requirements was a one-click installation.

doc/dev-guide/reconciler.md
nbalacha: There are multiple nodes and if multiple CRs are created at the same time, can we guarantee that all the nodes will process them in the same order?
If different nodes process the CRs in a different order, we cannot know which VGs can exist on which nodes. As the Storageclasses will have been created by then, potentially PVs can be created and then cleaning up becomes difficult. Hence I want to have a level of control on this.

travisn: The API server will notify controllers about the CRs in the order of creation, so I expect the ordering will be consistent. 

As discussed on the call, I'm ok to keep this behavior for now until we give this more time to bake. In 4.10 it's anyway simplified with a single VG supported. I also don't see upgrades being affected negatively if we do decide later to remove the VGs from the LVMCluster CR. The LVMVolumeGroup CRs would have already been created by the operator, and the operator could just ignore the VGs in the LVMCluster after that.

I can see if that if we have compelling validation error-checking scenarios for clusters with multiple VGs we could keep this design. But fundamentally I still recommend not including the VGs in the LVMCluster CR for these reasons:
1. Bundling multiple resources into a single CR and forcing it to be a singleton is not a common K8s pattern I would expect on CRDs. A volume group is a logical unit, and therefore each VG makes sense as a single CR instance.  
2. Validation must be done on the nodes anyway by the vgmanager where the devices can be fully validated. The operator just can't fully validate available devices.
3. Instead of assuming multiple VGs would cause conflicts as an error condition, allow the first VG to consume the device. The later VG will simply not find the available device. The later VG won't be in error, it just might not have devices available to fulfill storage requests.

nbalacha: As discussed on the call, I do not see any advantage to having the admin create the LVMVolumeGroup CRs. The actual interface exposed to the admin is the deviceClass - that is what we are internally mapping to lvm volume groups. We could decide to change this completely in the future.
The additional level of validation is something that makes things easier with setup especially when targeting hundreds of nodes. The node level validation done by the Vg manager is a different level of validation.

``` A volume group is a logical unit, and therefore each VG makes sense as a single CR instance.```
Yes, but that is not something the user needs to know. As far as they are concerned they get dynamic local storage and the exposing the lower layers to them is not really required. The LVMCluster CR will be fine tuned in future to provide this experience.

```Instead of assuming multiple VGs would cause conflicts as an error condition, allow the first VG to consume the device. The later VG will simply not find the available device. The later VG won't be in error, it just might not have devices available to fulfill storage requests.```

If there are failures in reconciliation for one VG or the reconcile requests are, for some reason, sent in a different order (network issues etc), this can lead to different VGs on different nodes. Since cleaning up is not easy, it makes more sense to prevent such conflicts as early as possible.

As discussed, this is the approach we are following now. Any further decisions will be based on user experience and feedback and will be taken up later.

File: doc/dev-guide/lvmo-units.md (hunk 4 above)
Comment (leelavg) [reply doc/dev-guide/lvmo-units.md:9]: @nbalacha do you want to see the mention of hostPath?

Comment (nbalacha) [reply doc/dev-guide/lvmo-units.md:9]: Yes, I think having the hostpath information here would be useful.

File: doc/dev-guide/reconciler.md (hunk 1 above)
Comment (nbalacha) [reply doc/dev-guide/reconciler.md:14]: We can look into that.

File: doc/dev-guide/topolvm-csi.md [hunk 6]
```diff
@@ -0,0 +1,35 @@
+# Topolvm CSI
+
+- LVM Operator deploys the Topolvm CSI which provides dynamic provisioning of
+  local storage.
+- Please refer to topolvm [docs][topolvm-docs] for more details on topolvm
+
+## CSI Driver
+
+- *csiDriver* reconcile unit deploys Topolvm CSIDriver resource
+- The LVM operator does not expose support for ephemeral volumes.
```
nbalacha, line 10: We can remove this sentence as the latest topolvm has removed ephemeral volumes support.

PR #103, opened 2022-02-04
Adding an alert mechanism into lvm-operator

aruniiird:
Adding an alert mixins framework into LVM Operator.
Makefile is modified to generate the prometheus alert yaml file.

Added 'VolumeGroupUsageAtThreshold' alert which will be triggered if
volumegroup usage goes beyond the threshold percentage (default set to 75%)

Signed-off-by: Arun Kumar Mohan <amohan@redhat.com>

### DIFF
--- a/config/prometheus/prometheus_rules.yaml
+++ b/config/prometheus/prometheus_rules.yaml
@@ -1,0 +1,26 @@
+"apiVersion": "monitoring.coreos.com/v1"
+"kind": "PrometheusRule"
+"metadata":
+  "name": "prometheus-lvmo-rules"
+"spec":
+  "groups":
+  - "name": "vg-alert.rules"
+    "rules":
+    - "alert": "VolumeGroupUsageAtThresholdNearFull"
+      "annotations":
+        "description": "VolumeGroup is nearing full. Data deletion or VolumeGroup expansion is required."
+        "message": "VolumeGroup {{ $labels.device_class }} utilization has crossed 75 % on node {{ $labels.node }}. Free up some space or expand the VolumeGroup."
+      "expr": |
+        (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > 0.75 and (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes <= 0.85
+      "for": "5m"
+      "labels":
+        "severity": "warning"
+    - "alert": "VolumeGroupUsageAtThresholdCritical"
+      "annotations":
+        "description": "VolumeGroup is critically full. Data deletion or VolumeGroup expansion is required."
+        "message": "VolumeGroup {{ $labels.device_class }} utilization has crossed 85 % on node {{ $labels.node }}. Free up some space or expand the VolumeGroup immediately."
+      "expr": |
+        (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > 0.85
+      "for": "5m"
+      "labels":
+        "severity": "critical"
--- a/monitoring/alerts/vgalerts.libsonnet
+++ b/monitoring/alerts/vgalerts.libsonnet
@@ -1,0 +1,39 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'vg-alert.rules',
+        rules: [
+          {
+            alert: 'VolumeGroupUsageAtThresholdNearFull',
+            expr: |||
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThresholdNearFull)0.2f and (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes <= %(vgUsageThresholdCritical)0.2f
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'VolumeGroup is nearing full. Data deletion or VolumeGroup expansion is required.',
+              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed %.0f %% on node {{ $labels.node }}. Free up some space or expand the VolumeGroup.' % ($._config.vgUsageThresholdNearFull*100),
+            },
+          },
+          {
+            alert: 'VolumeGroupUsageAtThresholdCritical',
+            expr: |||
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThresholdCritical)0.2f
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              description: 'VolumeGroup is critically full. Data deletion or VolumeGroup expansion is required.',
+              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed %.0f %% on node {{ $labels.node }}. Free up some space or expand the VolumeGroup immediately.' % ($._config.vgUsageThresholdCritical * 100),
+            },
+          },
+	],
+      },
+    ],
+  },
+}
--- a/monitoring/mixin.libsonnet
+++ b/monitoring/mixin.libsonnet
@@ -1,0 +1,14 @@
+(import 'config.libsonnet') +
+(import 'alerts/vgalerts.libsonnet') + {
+  prometheus+:: {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'PrometheusRule',
+        metadata: {
+          name: 'prometheus-lvmo-rules',
+        },
+        spec: {
+          groups: $.prometheusAlerts.groups,
+        },
+  }
+}
+
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,9 @@ build: generate fmt vet ## Build manager binary.
 build-vgmanager: generate fmt vet ## Build vg manager binary.
 	go build -o bin/vgmanager cmd/vgmanager/main.go
 
+build-prometheus-alert-rules: jsonnet monitoring/mixin.libsonnet monitoring/alerts/alerts.jsonnet monitoring/alerts/*.libsonnet
+	$(JSONNET) -S monitoring/alerts/alerts.jsonnet > config/prometheus/prometheus_rules.yaml
+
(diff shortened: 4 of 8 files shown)


nbalacha (dismissed):
Is it possible to just create a static YAML file for this?

nbalacha:
Please update the commit msg to match the commitlint requirements

iamniting (changes requested):
After discussing This with @aruniiird I understand that we can create/update the YAML files using the `jsonnet` tool via just changing the values in the `monitoring/config.libsonnet` file instead of changing the YAML file directly if we require any change in the future. Also, we can create different YAML files too using this tool for different use cases eg. Managed Services, snow, etc.

So the question to @nbalacha is do we really want to have that flexibility? if not we can create a static file too as you suggested.

If we are gonna stick to the same format we need to make sure that
- Make target also download `jsonnet` tool if not already present on the machine.
- We should use kustomize to use metadata from a different file and spec from a different file as this tool just create only spec and not metadata at all.
- We should also upload the generated YAML file.

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,28 @@
...
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: |||
+                'VolumeGroup usage has crossed the provided threshold of %(vgUsageThreshold)0.2f'
+              ||| %$._config,
+              description: 'VolumeGroup usage has grown up and crossed the provided thershold',
```
nbalacha, line 20:
```suggestion
              description: 'VolumeGroup usage has crossed the provided threshold',
```

monitoring/mixin.libsonnet
```diff
@@ -0,0 +1,15 @@
+(import 'config.libsonnet') +
+(import 'alerts/vgalerts.libsonnet') + {
+  prometheus+:: {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'PrometheusRule',
+        metadata: {
+          name: 'prometheus-vg-rules',
+          namespace: 'default',
```
nbalacha, line 8: namespace should not be default here. I think kustomize looks for system.

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,28 @@
...
+            expr: |||
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThreshold)0.2f
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: |||
+                'VolumeGroup usage has crossed the provided threshold of %(vgUsageThreshold)0.2f'
```
nbalacha, line 18: Please add the volume group and node name as well. I believe it is available in the topolvm metrics :

```curl -s 127.0.0.1:41685/metrics | grep -Ei 'topolvm_volumegroup_.*?_bytes\{'
topolvm_volumegroup_available_bytes{device_class="vg1",node="kube-node"} 4.790222323712e+12
topolvm_volumegroup_size_bytes{device_class="vg1",node="kube-node"} 4.800959741952e+12```

**Review by nbalacha — CHANGES_REQUESTED** [review]

File: monitoring/alerts/vgalerts.libsonnet (hunk 1 above)
Comment (iamniting) [reply monitoring/alerts/vgalerts.libsonnet]: Done

File: monitoring/mixin.libsonnet (hunk 2 above)
Comment (iamniting) [reply monitoring/mixin.libsonnet]: Done

File: monitoring/alerts/vgalerts.libsonnet (hunk 3 above)
Comment (pcuzner) [reply monitoring/alerts/vgalerts.libsonnet:18]: Apologies for the dumb question @nbalacha - but why are the labels called device_class and node? the content appears to be a vg name, the metric name is a volumegroup..and does the node refer to a hostname? If the labels are aligned to the schema in node-exporter for example, you can use promql more easily to merge the content.

Comment (iamniting) [reply monitoring/alerts/vgalerts.libsonnet]: @nbalacha done

File: monitoring/alerts/vgalerts.libsonnet [hunk 4]
```diff
@@ -0,0 +1,41 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'vg-alert.rules',
+        rules: [
+          {
+            alert: 'VolumeGroupUsageAtThresholdNearFull',
+            expr: |||
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThresholdNearFull)0.2f
```
nbalacha, line 10: This should be fired only if > 75% and < 85%

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,41 @@
...
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'VolumeGroup is nearing full. Data deletion or VolumeGroup expansion is required.',
+              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed 75% percent on node {{ $labels.node }}. Free up some space or expand the VolumeGroup.',
```
umangachapagain, line 18:
```suggestion
              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed %(vgUsageThresholdNearFull)0.2f percent on node {{ $labels.node }}. Free up some space or expand the VolumeGroup.',
```

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,41 @@
...
+            expr: |||
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThresholdCritical)0.2f
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              description: 'VolumeGroup is critically full. Data deletion or VolumeGroup expansion is required.',
+              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed 85% percent on node {{ $labels.node }}. Free up some space or expand the VolumeGroup immediately.',
```
umangachapagain, line 33:
```suggestion
              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed %(vgUsageThresholdCritical)0.2f percent on node {{ $labels.node }}. Free up some space or expand the VolumeGroup immediately.',
```

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,41 @@
...
+              severity_level: 'error',
```
umangachapagain, line 34:
```suggestion
```

monitoring/alerts/vgalerts.libsonnet
```diff
@@ -0,0 +1,41 @@
...
+              (topolvm_volumegroup_size_bytes - topolvm_volumegroup_available_bytes)/topolvm_volumegroup_size_bytes > %(vgUsageThresholdNearFull)0.2f
+            ||| % $._config,
+            'for': $._config.volumegroupUsageThresholdAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'VolumeGroup is nearing full. Data deletion or VolumeGroup expansion is required.',
+              message: 'VolumeGroup {{ $labels.device_class }} utilization has crossed 75% percent on node {{ $labels.node }}. Free up some space or expand the VolumeGroup.',
+              severity_level: 'warning',
```
umangachapagain, line 19:
```suggestion
```

umangachapagain (changes requested):

umangachapagain (commented):
Complexity of generating alerts using mixin seems to be more than the complexity of maintaining a single YAML file. So, for now I'd suggest using just a single static YAML file.
Once it gets fairly complex to manage, you can move to mixins or some other solutions.

monitoring/alerts/vgalerts.libsonnet
aruniiird: Done

monitoring/alerts/vgalerts.libsonnet
aruniiird: Done

monitoring/alerts/vgalerts.libsonnet
aruniiird: Done...

monitoring/alerts/vgalerts.libsonnet
aruniiird: Done

monitoring/alerts/vgalerts.libsonnet
aruniiird: Done

monitoring/alerts/vgalerts.libsonnet
nbalacha: @pcuzner , Not dumb at all. These metrics are exposed by the topolvm csi driver and they use something called a deviceClass  which internally maps to the VG. The labels are from topolvm.

monitoring/mixin.libsonnet
```diff
@@ -0,0 +1,14 @@
+(import 'config.libsonnet') +
+(import 'alerts/vgalerts.libsonnet') + {
+  prometheus+:: {
+        apiVersion: 'monitoring.coreos.com/v1',
+        kind: 'PrometheusRule',
+        metadata: {
+          name: 'prometheus-vg-rules',
```
nbalacha, line 7:
```suggestion
          name: 'prometheus-lvmo-rules',
```

config/prometheus/prometheus_rules.yaml
```diff
@@ -0,0 +1,26 @@
+"apiVersion": "monitoring.coreos.com/v1"
+"kind": "PrometheusRule"
+"metadata":
+  "name": "prometheus-vg-rules"
```
nbalacha, line 4:
```suggestion
  "name": "prometheus-lvmo-rules"
```

nbalacha (changes requested):

monitoring/mixin.libsonnet
iamniting: done

config/prometheus/prometheus_rules.yaml
iamniting: done

nbalacha:
> Complexity of generating alerts using mixin seems to be more than the complexity of maintaining a single YAML file. So, for now I'd suggest using just a single static YAML file. Once it gets fairly complex to manage, you can move to mixins or some other solutions.

We did discuss whether we should just use a static YAML and decided to put that effort in now in order to make it easier in the future.

umangachapagain (approved):

nbalacha (approved):

Issue #105, opened 2022-02-07
No news disks are added to the LVMCluster/LVMvolumegroupnodestatuses


jgato:
I had one SNO with an LVMCluster created to manage the VG vg1 (/dev/nvme0n1, /dev/nvme1n1, /dev/sda). I wanted to test the addition of new disks. So, I rebooted and created some more disks from the raid I have in the server. 

After creating some more disks the SNO is rebooted and:
 * The new disks (/dev/sdc, /dev/sdd) have been added to the vg1 in the node:
```
[root@master-0 core]# vgs
  VG  #PV #LV #SN Attr   VSize VFree 
  vg1   5   4   0 wz--n- 6.00t <6.00t
[root@master-0 core]# pvs
  PV           VG  Fmt  Attr PSize   PFree  
  /dev/nvme0n1 vg1 lvm2 a--  745.21g 745.21g
  /dev/nvme1n1 vg1 lvm2 a--  745.21g 745.21g
  /dev/sda     vg1 lvm2 a--   <2.73t   2.72t
  /dev/sdc     vg1 lvm2 a--  931.48g 931.48g
  /dev/sde     vg1 lvm2 a--  931.48g 931.48g

```
 * This new disks are not recognized by the LVMCluster/LVMvolumegroupnodestatuses:
```
$ oc get lvmvolumegroupnodestatuses -o yaml
apiVersion: v1
items:
- apiVersion: lvm.topolvm.io/v1alpha1
  kind: LVMVolumeGroupNodeStatus
  metadata:
    creationTimestamp: "2022-02-05T18:58:01Z"
    generation: 1
    name: master-0.apollo2.hpecloud.org
    namespace: lvm-operator-system
    resourceVersion: "6276721"
    uid: b6cc0395-e17c-4116-941f-9b95b8c4ed82
  spec:
    nodeStatus:
    - devices:
      - /dev/nvme0n1
      - /dev/nvme1n1
      - /dev/sda
      name: vg1
      status: Ready
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
```
$ oc get lvmcluster  lvmcluster-sample -o yaml
apiVersion: lvm.topolvm.io/v1alpha1
kind: LVMCluster
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #110, opened 2022-02-15
fix: updates lvmvolumegroupnodestatus for vg

nbalacha:
Fixes a bug where the lvmvolumegroupnodestatus was not updated correctly
for an existing VG. When adding a new device and restarting the
vg-manager pod, the device is added to the LVM VG but is not updated in
the lvmvolumegroupnodestatus or the LVMCluster status.

fixes: #105

### DIFF
--- a/pkg/vgmanager/vgmanager_controller.go
+++ b/pkg/vgmanager/vgmanager_controller.go
@@ -318,12 +318,10 @@ func setStatus(status *lvmv1alpha1.VGStatus, instance *lvmv1alpha1.LVMVolumeGrou
 	found := false
 
 	vgStatuses := instance.Spec.LVMVGStatus
-	for _, vgStatus := range vgStatuses {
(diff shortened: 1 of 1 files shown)


pkg/vgmanager/vgmanager_controller.go
```diff
@@ -318,12 +318,9 @@ func setStatus(status *lvmv1alpha1.VGStatus, instance *lvmv1alpha1.LVMVolumeGrou
 	found := false
 
 	vgStatuses := instance.Spec.LVMVGStatus
-	for _, vgStatus := range vgStatuses {
+	for i, vgStatus := range vgStatuses {
 		if vgStatus.Name == status.Name {
-			found = true
```
leelavg, line 323: - not toggling `found` may create a new vgstatus as per condition in line 331
- so `found` should still be set to `true` 🤔

leelavg (approved):
/lgtm

PR #111, opened 2022-02-15
fix: updates prebuilt img location in doc

nbalacha:
Updated the install instructions to point to the lvm-operator
image in quay.io/ocs-dev.


nbalacha:
A better way would be to just link it to the doc which describes how to build the operator. That can be taken up in a separate PR.

leelavg (approved):
/lgtm

PR #114, opened 2022-02-17
fix: renamed storageclass to remove topolvm

nbalacha:
Replaced "topolvm-" with "odf-lvm" in the storageclass name.

### DIFF
--- a/controllers/topolvm_storageclass.go
+++ b/controllers/topolvm_storageclass.go
@@ -51,7 +51,7 @@ func (s topolvmStorageClass) ensureDeleted(r *LVMClusterReconciler, ctx context.
 	// delete the corresponding storage class
 	for _, deviceClass := range lvmCluster.Spec.DeviceClasses {
 		sc := &storagev1.StorageClass{}
-		scName := fmt.Sprintf("topolvm-%s", deviceClass.Name)
+		scName := fmt.Sprintf("odf-lvm-%s", deviceClass.Name)
 		err := r.Client.Get(ctx, types.NamespacedName{Name: scName}, sc)
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


Yuggupta27 (commented):
Do we need to update [here](https://github.com/red-hat-storage/lvm-operator/blob/1ed0e13d84f7f1b953c444f1655219878c19af5a/controllers/topolvm_storageclass.go#L16) too?

controllers/topolvm_storageclass.go
```diff
@@ -93,7 +93,7 @@ func getTopolvmStorageClasses(lvmCluster *lvmv1alpha1.LVMCluster) []*storagev1.S
 	for _, deviceClass := range lvmCluster.Spec.DeviceClasses {
 		storageClass := &storagev1.StorageClass{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: fmt.Sprintf("topolvm-%s", deviceClass.Name),
+				Name: fmt.Sprintf("odf-lvm-%s", deviceClass.Name),
```
Madhu-1, line 96: @nbalacha not sure it a problem for upgrades update of storageclass name is not supported we need to delete and recreate it?

nbalacha: Upgrades are not in the picture as yet as this is the first downstream release.

nbalacha:
> Do we need to update [here](https://github.com/red-hat-storage/lvm-operator/blob/1ed0e13d84f7f1b953c444f1655219878c19af5a/controllers/topolvm_storageclass.go#L16) too?

Not required. This is the name of the resource manager, not the storageclass.

controllers/topolvm_storageclass.go
leelavg, line 96: - with this change, we are make the project more downstream-ish, as we already have code to determine k8s environment (openshift or other), how about we use that info to decide on `odf-lvm-` vs `topolvm-`?

nbalacha: I think a more generic sc name is probably fine for both upstream and downstream.

leelavg: ack, so we'll proceed with this name for now and later change it to generic then.

leelavg (approved):
/lgtm

PR #128, opened 2022-03-09
build: add nameprefix to the controller manager

iamniting:

https://bugzilla.redhat.com/show_bug.cgi?id=2047173

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,8 @@ SHELL = /usr/bin/env bash -o pipefail
 IMAGE_BUILD_CMD ?= $(shell command -v docker 2>&1 >/dev/null && echo docker || echo podman)
 
 
+MANAGER_NAME_PREFIX ?= lvm-operator-
+
 all: build
 
 ##@ General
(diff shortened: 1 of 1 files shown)


leelavg (approved):
/lgtm

nbalacha (commented):
This works for the controller-manager deployment and pod. Is there any way to do this for the serviceaccount as well?

leelavg:
- I suppose you mean the controller-manager service account
- If that's the case we may need to write a patch file targeting only this serviceAccount and corresponding roles/roleBindings/deployment as I believe we can't set this via kustomize cli
- Needs to be investigated further

iamniting:
> This works for the controller-manager deployment and pod. Is there any way to do this for the serviceaccount as well?

We could do that, But eventually, it will bring much more complexity in the kustomize and makefile because of rbacs and all, I would suggest that we just change the name of the service account to something appropriate so that it won't conflict with something else.

nbalacha (approved):

PR #133, opened 2022-03-22
fix: implements the nodeSelector check

nbalacha:
This commit implements the nodeSelector check in
vgmanager.

### DIFF
--- a/pkg/vgmanager/vgmanager_controller.go
+++ b/pkg/vgmanager/vgmanager_controller.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	corev1helper "k8s.io/component-helpers/scheduling/corev1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -90,12 +91,11 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		return ctrl.Result{}, err
 	}
 
-	//TODO: actually check the node against the nodeSelector.
-	node := &corev1.Node{}
-	nodeMatches, err := NodeSelectorMatchesNodeLabels(node, volumeGroup.Spec.NodeSelector)
+	//Check if the VG nodeSelector matches the labels on this node
+	nodeMatches, err := r.matchesThisNode(ctx, volumeGroup.Spec.NodeSelector)
 	if err != nil {
 		r.Log.Error(err, "failed to match nodeSelector to node labels", "VGName", volumeGroup.Name)
-		return reconcileAgain, err
+		return ctrl.Result{}, err
 	}
 
 	if !nodeMatches {
@@ -380,4 +380,13 @@ func (r *VGReconciler) getNewNodeStatus(status *lvmv1alpha1.VGStatus) *lvmv1alph
 	}
 	setStatus(status, vgNodeStatus)
 	return vgNodeStatus
+}
+
+func (r *VGReconciler) matchesThisNode(ctx context.Context, selector *corev1.NodeSelector) (bool, error) {
+	node := &corev1.Node{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: r.NodeName}, node)
+	if err != nil {
+		return false, err
+	}
+	return NodeSelectorMatchesNodeLabels(node, selector)
 }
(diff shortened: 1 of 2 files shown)


pkg/vgmanager/vgmanager_controller.go
```diff
@@ -90,12 +91,11 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 		return ctrl.Result{}, err
 	}
 
-	//TODO: actually check the node against the nodeSelector.
-	node := &corev1.Node{}
-	nodeMatches, err := NodeSelectorMatchesNodeLabels(node, volumeGroup.Spec.NodeSelector)
+	//Does this node need to process it?
```
sp98, line 94: can you elaborate a bit more?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -381,3 +381,21 @@ func (r *VGReconciler) getNewNodeStatus(status *lvmv1alpha1.VGStatus) *lvmv1alph
...
+	err := r.Client.Get(ctx, types.NamespacedName{Name: r.NodeName}, node)
+	if err != nil {
+		return false, err
+	}
+	nodeMatches, err := NodeSelectorMatchesNodeLabels(node, selector)
+	if err != nil {
+		return false, err
+	}
+
+	if !nodeMatches {
```
sp98, line 397: This check might not be needed. Function can return `nodeMatches` variable.

pkg/vgmanager/vgmanager_controller.go
nbalacha: If the nodeSelector for the VG does not match the node the vgmanager is running on, it does not need to do anything.

pkg/vgmanager/vgmanager_controller.go
nbalacha: done.

pkg/vgmanager/vgmanager_controller.go
sp98: ok.  The comment is framed as a question.  Should it rather be a statement.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -381,3 +382,14 @@ func (r *VGReconciler) getNewNodeStatus(status *lvmv1alpha1.VGStatus) *lvmv1alph
...
 }
+
+func (r *VGReconciler) matchesMyNode(ctx context.Context, selector *corev1.NodeSelector) (bool, error) {
+
+	node := &corev1.Node{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: r.NodeName}, node)
+	if err != nil {
+		return false, err
+	}
+	nodeMatches, err := NodeSelectorMatchesNodeLabels(node, selector)
```
sp98, line 393: maybe just do 
` return NodeSelectorMatchesNodeLabels(node, selector)`

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -381,3 +382,14 @@ func (r *VGReconciler) getNewNodeStatus(status *lvmv1alpha1.VGStatus) *lvmv1alph
 	setStatus(status, vgNodeStatus)
 	return vgNodeStatus
 }
+
+func (r *VGReconciler) matchesMyNode(ctx context.Context, selector *corev1.NodeSelector) (bool, error) {
```
sp98, line 386: need a better name than `matchesMynode`.

leelavg, line 386: - maybe we'll not be using this function again and considering it's quite small, how about in-lining it in the call site itself?

nbalacha: I think that is best left to the compiler.

nbalacha: Suggestions?

leelavg: ah! got it.

sp98: I also agree with Leela here. We can remove the `matchesMyNode` function.

sp98: `matchesNode`.   But I also agree with Leela's comment that we can remove this function.

nbalacha: I prefer having functions for readability.

pkg/vgmanager/vgmanager_controller.go
nbalacha: done

pkg/vgmanager/vgmanager_controller.go
nbalacha: I disagree - functions provide readability.

leelavg (approved):
/lgtm

PR #140, opened 2022-04-01
fix: fix ensureDeleted of openshiftSccs via returning an error

iamniting:
The func does not return an error if there is an error and always return
nil. Fix the same via returning and error if there non-expected failure.


iamniting:
> Code changes seem fine. Please check the golang lint ci run to see what went wrong.

golangci is fixed now with another PR from Leela.

PR #142, opened 2022-04-04
fix: add node related commands to must-gather

agarwal-mudit:
Add node related commands to must-gather output

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>

### DIFF
--- a/must-gather/collection-scripts/gather_node_info
+++ b/must-gather/collection-scripts/gather_node_info
@@ -1,0 +1,46 @@
+#!/bin/bash
+# Expect base collection path as an argument
+BASE_COLLECTION_PATH=$1
+
+# Use PWD as base path if no argument is passed
+if [ "${BASE_COLLECTION_PATH}" = "" ]; then
+    BASE_COLLECTION_PATH=$(pwd)
+fi
+
+# Command List
+commands_get=()
+
+# lvm commands
+commands_get+=("lvs -a")
+commands_get+=("lvdisplay")
+commands_get+=("vgs")
+commands_get+=("vgdisplay")
+commands_get+=("pvs")
+commands_get+=("pvdisplay")
+commands_get+=("lvm version")
+commands_get+=("cat /etc/topolvm/lvmd.yaml")
+
+# collection path for lvm commands
+COMMAND_OUTPUT_FILE="${BASE_COLLECTION_PATH}/node_info"
+
+printf "collecting output of node related commands\n" | tee -a "${BASE_COLLECTION_PATH}/gather-debug.log"
+
+# Create the command, push all the commands to a temp file
+COMMAND_INPUT_FILE="${BASE_COLLECTION_PATH}/temp_file"
+printf "chroot /host" >> "${COMMAND_INPUT_FILE}"
+
+STRING="\\\n"
+for command_get in "${commands_get[@]}"; do
+    {
+        printf "\nprintf \"output of command: %s\"\n" "${command_get}"
+        printf "printf \"%s\"\n" "${STRING}"
+        printf "%s\n" "${command_get}"
+        printf "printf \"%s\"\n" "${STRING}"
+ } >> "${COMMAND_INPUT_FILE}"
+done
+
+node=$(oc get nodes -oname)
+
+{ timeout 120 cat "${COMMAND_INPUT_FILE}" | oc debug  -q "${node}" >> "${COMMAND_OUTPUT_FILE}"; }
+
+rm "${COMMAND_INPUT_FILE}"
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -12,6 +12,7 @@ printf "collection started at: %s \n" "${START_TIME}" >> ${BASE_COLLECTION_PATH}
 
 
 # Call other gather scripts
+gather_node_info ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
 gather_namespaced_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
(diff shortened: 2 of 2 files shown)


agarwal-mudit:
Its a bit ugly but this is the best I can find as of now, attaching the output file as well.
[node_info.txt](https://github.com/red-hat-storage/lvm-operator/files/8408978/node_info.txt)

must-gather/collection-scripts/gather_node_info
```diff
@@ -0,0 +1,45 @@
...
+STRING="\\\n"
+for command_get in "${commands_get[@]}"; do
+     # shellcheck disable=SC2129
+     printf "\nprintf \"output of command: %s\"\n" "${command_get}" >> "${COMMAND_INPUT_FILE}"
+     printf "printf \"%s\"\n" "${STRING}" >> "${COMMAND_INPUT_FILE}"
+     printf "%s\n" "${command_get}" >> "${COMMAND_INPUT_FILE}"
+     printf "printf \"%s\"\n" "${STRING}" >> "${COMMAND_INPUT_FILE}"
+done
+
+node=$(oc get nodes --no-headers | awk '/\yReady\y/{print $1}')
```
leelavg, line 41: - maybe `oc get nodes -oname` will be good enough?

must-gather/collection-scripts/gather_node_info
```diff
@@ -0,0 +1,45 @@
...
+
+printf "collecting output of node related commands" | tee -a "${BASE_COLLECTION_PATH}/gather-debug.log"
+
+# Create the command, push all the commands to a temp file
+COMMAND_INPUT_FILE="${BASE_COLLECTION_PATH}/temp_file"
+printf "chroot /host" >> "${COMMAND_INPUT_FILE}"
+
+STRING="\\\n"
+for command_get in "${commands_get[@]}"; do
+     # shellcheck disable=SC2129
```
leelavg, line 34: - SC2129 seems to be good, any issue conforming to that?
- this seems to be working for me
```    
 {
      printf "\nprintf \"output of command: %s\"\n" "${command_get}" 
      printf "printf \"%s\"\n" "${STRING}" 
      printf "%s\n" "${command_get}" 
      printf "printf \"%s\"\n" "${STRING}" 

 } >> "${COMMAND_INPUT_FILE}"
```

leelavg (commented):
> Its a bit ugly but this is the best I can find as of now,
- not sure what you mean by this, do we want to conform to any pattern as the collected info seems good enough to me.

must-gather/collection-scripts/gather_node_info
agarwal-mudit: Done

must-gather/collection-scripts/gather_node_info
agarwal-mudit: Done, thanks for the suggestion

agarwal-mudit:
> > Its a bit ugly but this is the best I can find as of now,
> 
>     * not sure what you mean by this, do we want to conform to any pattern as the collected info seems good enough to me.

I meant that we have to use oc debug command and because of that I have to put everything in a file and then execute that file as a single command. Couldn't think of any other solution, I need to check with LVM developer also.

leelavg:
> because of that I have to put everything in a file 
- The other way of doing this is a bit involved and so I believe this is good enough, it's like you need to run `tail -f /dev/null` and bg the process
- Then you'll have a debug pod running and you could run commands from cli in this pod as any other pod however as part of cleanup, you need to capture the PID of bg process and send a kill signal, pros will be no need to create cmd file

> I need to check with LVM developer also.
- ack, IMHO we could just add `-a` to `lvm lvs` so that it'll show metadata & data vols for vols on thinpool (thin vols and snapshots)

> we have to use oc debug command
- one tip would be to use `-q` with `oc debug` so that all the debug info will be suppressed

agarwal-mudit:
@leelavg added -a and -q as per the last comment, PTAL

leelavg:
- lgtm, if I approve this'll get merged, I'll await @nbalacha review as well.

must-gather/collection-scripts/gather_node_info
```diff
@@ -0,0 +1,46 @@
...
+# Use PWD as base path if no argument is passed
+if [ "${BASE_COLLECTION_PATH}" = "" ]; then
+    BASE_COLLECTION_PATH=$(pwd)
+fi
+
+# Command List
+commands_get=()
+
+# lvm commands
+commands_get+=("lvm lvs -a")
```
nbalacha, line 14: we don't need to call lvm each time. For instance, just lvs, vgs etc will work. Why are we including lvm as well? lvm apparently opens the lvm shell.

must-gather/collection-scripts/gather_node_info
```diff
@@ -0,0 +1,46 @@
...
+commands_get+=("lvm lvdisplay")
+commands_get+=("lvm vgs")
+commands_get+=("lvm vgdisplay")
+commands_get+=("lvm pvs")
+commands_get+=("lvm pvdisplay")
+commands_get+=("lvm version")
```
nbalacha, line 20: This is the only command that needs lvm in the command.

must-gather/collection-scripts/gather_node_info
nbalacha: Lets check with the LVM dev if this could have any unwanted effects. Otherwise I think this is fine.

agarwal-mudit: I will check with him along with the other things, but as far as I have noticed running only "lvm" opens a lvm shell otherwise not. But yeah, all the commands work without prefixing lvm so I can remove that.

nbalacha (approved):

PR #144, opened 2022-04-05
e2e: added make target to install ginkgo

riya-singhal31:
Added make target to install ginkgo and gomega

Signed-off-by: riya-singhal31 <riyasinghalji@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,6 +204,11 @@ JSONNET = $(shell pwd)/bin/jsonnet
 jsonnet: ## Download jsonnet locally if necessary.
 	$(call go-get-tool,$(JSONNET),github.com/google/go-jsonnet/cmd/jsonnet@latest)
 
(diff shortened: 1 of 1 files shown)


nbalacha:
1. I don't see the e2e-test target in this file
2. The ginkgo target must be called in the e2e-test target

Makefile
```diff
@@ -310,6 +315,6 @@ LVM_OPERATOR_UNINSTALL ?= true
 # Build and run lvm tests.
 e2e-test:
```
nbalacha, line 316:
```suggestion
e2e-test: ginkgo
```

riya-singhal31: Thanks Nithya for the suggestion,
It is working and is able to install ginkgo if not already installed.
It creates a ginkgo file inside the bin package if we already dont have that.

nbalacha (approved):

PR #146, opened 2022-04-06
feat: cleanup lvm resources on delete

nbalacha:
The vgmanager will now remove the vg and the lvmd.conf
file when the LVMCluster is deleted.

### DIFF
--- a/controllers/lvm_volumegroup.go
+++ b/controllers/lvm_volumegroup.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -27,7 +28,8 @@ import (
 )
 
 const (
-	lvmVGName = "lvmvg-manager"
+	lvmVGName      = "lvmvg-manager"
+	lvmvgFinalizer = "lvm.openshift.io/lvmvolumegroup"
 )
 
 type lvmVG struct{}
@@ -61,6 +63,7 @@ func (c lvmVG) ensureDeleted(r *LVMClusterReconciler, ctx context.Context, lvmCl
 
 	existingLvmVg := &lvmv1alpha1.LVMVolumeGroup{}
 	vgcrs := c.getLvmVolumeGroups(r, lvmCluster)
+	allVGsDeleted := true
 
 	for _, vgcr := range vgcrs {
 		err := r.Client.Get(ctx, types.NamespacedName{Name: vgcr.Name, Namespace: vgcr.Namespace}, existingLvmVg)
@@ -82,8 +85,25 @@ func (c lvmVG) ensureDeleted(r *LVMClusterReconciler, ctx context.Context, lvmCl
 				return err
 			} else {
 				r.Log.Info("initiated LVMVolumeGroup deletion", "name", existingLvmVg.Name)
+				allVGsDeleted = false
+			}
+		} else {
+			// Has the VG been cleaned up on all hosts?
+			exists := doesVGExistOnHosts(existingLvmVg.Name, lvmCluster)
+			if !exists {
+				// Remove finalizer
+				cutil.RemoveFinalizer(existingLvmVg, lvmvgFinalizer)
+				err = r.Client.Update(ctx, existingLvmVg)
+				if err != nil {
+					return err
+				}
+			} else {
+				allVGsDeleted = false
 			}
 		}
+	}
+	if !allVGsDeleted {
+		return fmt.Errorf("waiting for all VGs to be deleted")
 	}
 	return nil
 }
@@ -103,6 +123,9 @@ func (c lvmVG) getLvmVolumeGroups(r *LVMClusterReconciler, instance *lvmv1alpha1
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      deviceClass.Name,
 				Namespace: r.Namespace,
+				Finalizers: []string{
+					lvmvgFinalizer,
+				},
 			},
 			Spec: lvmv1alpha1.LVMVolumeGroupSpec{
 				NodeSelector:   deviceClass.NodeSelector,
@@ -112,4 +135,15 @@ func (c lvmVG) getLvmVolumeGroups(r *LVMClusterReconciler, instance *lvmv1alpha1
 		lvmVolumeGroups = append(lvmVolumeGroups, lvmVolumeGroup)
 	}
 	return lvmVolumeGroups
+}
+
+func doesVGExistOnHosts(volumeGroup string, instance *lvmv1alpha1.LVMCluster) bool {
+
+	dcStatuses := instance.Status.DeviceClassStatuses
+	for _, dc := range dcStatuses {
+		if dc.Name == volumeGroup {
+			return true
+		}
+	}
+	return false
 }
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -29,6 +30,7 @@ import (
 	"github.com/go-logr/logr"
 	secv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+	topolvmv1 "github.com/topolvm/topolvm/api/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -74,6 +76,7 @@ type LVMClusterReconciler struct {
 //+kubebuilder:rbac:groups=lvm.topolvm.io,resources=lvmvolumegroupnodestatuses/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=lvm.topolvm.io,resources=lvmvolumegroupnodestatuses/finalizers,verbs=update
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;create;update;delete
+//+kubebuilder:rbac:groups=topolvm.cybozu.com,resources=logicalvolumes,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
@@ -115,14 +118,12 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	result, reconcileError := r.reconcile(ctx, lvmCluster)
 
 	statusError := r.updateLVMClusterStatus(ctx, lvmCluster)
-	if statusError != nil {
-		r.Log.Error(statusError, "failed to update VG Node status")
-	}
 
 	// Reconcile errors have higher priority than status update errors
 	if reconcileError != nil {
 		return result, reconcileError
-	} else if statusError != nil && errors.IsNotFound(statusError) {
+	} else if statusError != nil && !errors.IsNotFound(statusError) {
+		r.Log.Error(statusError, "failed to update LVMCluster status")
 		return result, statusError
 	} else {
 		return result, nil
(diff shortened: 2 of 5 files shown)


controllers/lvm_volumegroup.go
```diff
@@ -27,7 +28,8 @@ import (
 )
 
 const (
-	lvmVGName = "lvmvg-manager"
+	lvmVGName      = "lvmvg-manager"
+	lvmvgFinalizer = "lvm.openshift.com/volumegroup"
```
leelavg, line 32: - minor: i think `openshift.io` is used more often

controllers/lvmcluster_controller.go
```diff
@@ -115,14 +118,15 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	result, reconcileError := r.reconcile(ctx, lvmCluster)
 
 	statusError := r.updateLVMClusterStatus(ctx, lvmCluster)
-	if statusError != nil {
+	/* 	if statusError != nil {
```
leelavg, line 121: commented out block can be removed?

controllers/lvmcluster_controller.go
```diff
@@ -348,3 +338,47 @@ func (r *LVMClusterReconciler) getRunningPodImage(ctx context.Context) error {
 
 	return nil
 }
+
+func (r *LVMClusterReconciler) logicalVolumesExist(ctx context.Context, instance *lvmv1alpha1.LVMCluster) (bool, error) {
+
+	logicalVolumeList := &topolvmv1.LogicalVolumeList{}
+
+	if err := r.Client.List(ctx, logicalVolumeList); err != nil {
```
leelavg, line 346:
```suggestion
	if err := r.Client.List(ctx, logicalVolumeList, &client.ListOptions{Limit: int64(1)}); err != nil {
```

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -182,8 +188,10 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
 			status.Status = lvmv1alpha1.VGStatusDegraded
 			status.Reason = "VGExtendFailed"
 		}
-		if statuserr := r.updateStatus(ctx, status, volumeGroup); statuserr != nil {
+
+		if statuserr := r.updateStatus(ctx, lvmdConfig); statuserr != nil {
 			r.Log.Error(statuserr, "failed to update status", "VGName", volumeGroup.Name)
+			return reconcileAgain, nil
```
leelavg, line 194: - shouldn't `nil` be `statuserr` here?
- same comment in this file

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -208,11 +216,11 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
 		r.Log.Info("updated lvmd config", "VGName", volumeGroup.Name)
 	}
 
-	volGrpHostInfo, err := GetVolumeGroup(r.executor, volumeGroup.Name)
+	// volGrpHostInfo, err := GetVolumeGroup(r.executor, volumeGroup.Name)
```
leelavg, line 219: - this wasn't being used even before this PR?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -413,3 +423,67 @@ func (r *VGReconciler) getMatchingDevicesForVG(volumeGroup *lvmv1alpha1.LVMVolum
...
+		},
+		Spec: lvmv1alpha1.LVMVolumeGroupNodeStatusSpec{
+			LVMVGStatus: statusarr,
+		},
+	}
+
+	return vgNodeStatus, nil
+}
+
+func (r *VGReconciler) updateStatus(ctx context.Context, lvmdConfig *lvmdCMD.Config) error {
```
leelavg, line 459: - I don't see `lvmdConfig` is being used directly in this func or other funcs being called in this func
- If it's intentional we don't need this param at all?
- I was thinking this can be used to filter the vgs that lvmo is managing?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -413,3 +423,67 @@ func (r *VGReconciler) getMatchingDevicesForVG(volumeGroup *lvmv1alpha1.LVMVolum
...
+
+	result, err := ctrl.CreateOrUpdate(ctx, r.Client, nodeStatus, func() error {
+		nodeStatus.Spec.LVMVGStatus = vgNodeStatus.Spec.LVMVGStatus
+		return nil
+	})
+
+	if err != nil {
+		r.Log.Error(err, "failed to create or update lvmvolumegroupnodestatus", "name", vgNodeStatus.Name)
+		return err
+	} else if result != controllerutil.OperationResultNone {
```
leelavg, line 483: any reason for not checking against `OperationResultUpdated`

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -226,6 +234,71 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
...
+		return nil
+	}
+
+	// Check if volume group exists
+	vg, err := GetVolumeGroup(r.executor, volumeGroup.Name)
+	if err != nil {
+		if err != ErrVolumeGroupNotFound {
+			return fmt.Errorf("failed to get volume group. %v, %v", volumeGroup.Name, err)
+		}
+		return nil
```
leelavg, line 268: - why can't directly `return err`?
- if not above, why `return nil` in line 268?

leelavg (commented):
- there are some commented out log lines, can those be removed?

controllers/lvmcluster_controller.go
nbalacha: What is the advantage of this?

sp98 (changes requested):

controllers/lvm_volumegroup.go
```diff
@@ -113,3 +136,14 @@ func (c lvmVG) getLvmVolumeGroups(r *LVMClusterReconciler, instance *lvmv1alpha1
 	}
 	return lvmVolumeGroups
 }
+
+func doesVGExistOnHosts(ctx context.Context, volumeGroup string, instance *lvmv1alpha1.LVMCluster) (bool, error) {
```
leelavg, line 140: - it seems `error` is always nil and can be removed
- not a time limited func so `ctx` can be removed as well

controllers/lvm_volumegroup.go
```diff
@@ -113,3 +136,14 @@ func (c lvmVG) getLvmVolumeGroups(r *LVMClusterReconciler, instance *lvmv1alpha1
...
+
+	dcStatuses := instance.Status.DeviceClassStatuses
+	for _, dc := range dcStatuses {
+		if dc.Name == volumeGroup {
```
leelavg, line 144: - only for my understanding, we are finding whether VG exists or not by means of status of lvmcluster cr which inturn gets the info from groupnodestatus cr right?
- is it possible to club volumegroup and groupnodestatus crd? (if past discussions were a no, then nntr)

controllers/lvmcluster_controller.go
leelavg: - we are checking the existence of logicalvolume crs so why not limit the request to list only one item

pkg/vgmanager/vgmanager_controller.go
leelavg: ack, makes sense, we can use tags then if that's still an options, maybe in another pr.

pkg/vgmanager/vgmanager_controller.go
leelavg: ack, I read the log `lvmvolumegroupnodestatus modified` and thought this condition is checking solely the modification.

controllers/lvm_volumegroup.go
nbalacha: > * only for my understanding, we are finding whether VG exists or not by means of status of lvmcluster cr which inturn gets the info from groupnodestatus cr right?
Correct
> * is it possible to club volumegroup and groupnodestatus crd? (if past discussions were a no, then nntr)
Can you elaborate?

controllers/lvmcluster_controller.go
nbalacha: Does it make any difference from a performance point of view?

controllers/lvm_volumegroup.go
```diff
@@ -82,9 +85,26 @@ func (c lvmVG) ensureDeleted(r *LVMClusterReconciler, ctx context.Context, lvmCl
 				return err
 			} else {
 				r.Log.Info("initiated LVMVolumeGroup deletion", "name", existingLvmVg.Name)
+				allVGsDeleted = false
+			}
+		} else {
+			// Has the VG been cleaned up on all hosts?
+			exists, _ := doesVGExistOnHosts(ctx, existingLvmVg.Name, lvmCluster)
+			if !exists {
+				//remove finalizer
```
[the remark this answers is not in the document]
nbalacha: done

controllers/lvmcluster_controller.go
```diff
@@ -134,32 +138,18 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
-					r.Log.Error(err, "failed cleaning up", "resource", unit.getName())
-					return ctrl.Result{}, err
-				}
-			}
-			controllerutil.RemoveFinalizer(instance, lvmClusterFinalizer)
-			if err := r.Client.Update(context.TODO(), instance); err != nil {
-				r.Log.Info("failed to remove finalizer from LvmCluster")
-				return reconcile.Result{}, err
-			}
+		//Check for existing LogicalVolumes
```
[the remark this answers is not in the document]
nbalacha: done

controllers/lvmcluster_controller.go
```diff
@@ -134,32 +138,18 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
+		lvsExist, err := r.logicalVolumesExist(ctx, instance)
+		if err != nil {
+			r.Log.Error(err, "failed to get LogicalVolumeList")
```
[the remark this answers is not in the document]
nbalacha: done

controllers/lvm_volumegroup.go
nbalacha: done

controllers/lvmcluster_controller.go
nbalacha: done

pkg/vgmanager/vgmanager_controller.go
nbalacha: removed.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -226,6 +234,71 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
 	return ctrl.Result{RequeueAfter: requeueAfter}, nil
 }
 
+func (r *VGReconciler) processDelete(ctx context.Context, volumeGroup *lvmv1alpha1.LVMVolumeGroup) error {
+
+	//Read the lvmd config file
```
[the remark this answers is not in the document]
nbalacha: done

controllers/lvm_volumegroup.go
nbalacha: removed error.

pkg/vgmanager/vgmanager_controller.go
leelavg: > ErrVolumeGroupNotFound , there is nothing more to be done.
- ya, missed this detail

controllers/lvm_volumegroup.go
leelavg: > Can you elaborate?
- not related to this PR, can be taken up separately
- in this single function either directly or indirectly we are using all the three CRs with some duplicate info, even though we discussed in length about the current state, if possible we want to re-look in reducing the duplicate info there by reducing num of CRs

controllers/lvmcluster_controller.go
leelavg: - leaving this then, even though we might be reading from cache I think limiting entries doesn't harm anything for the operation that we are doing in this func

controllers/lvmcluster_controller.go
```diff
@@ -115,14 +118,12 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	statusError := r.updateLVMClusterStatus(ctx, lvmCluster)
-	if statusError != nil {
-		r.Log.Error(statusError, "failed to update VG Node status")
-	}
 
 	// Reconcile errors have higher priority than status update errors
 	if reconcileError != nil {
 		return result, reconcileError
-	} else if statusError != nil && errors.IsNotFound(statusError) {
+	} else if statusError != nil && !errors.IsNotFound(statusError) {
```
Yuggupta27, line 125: Do we don't need to return errors in the case of `NotFound` scenarios now?

leelavg (approved):
/lgtm

controllers/lvmcluster_controller.go
nbalacha: I think this was originally a mistake (as Nitin pointed out).

Issue #109, opened 2022-02-15
lsblk detecting my mini sas hd connected drives as removable


deanpeterson:
vg-manager logs:

```{"level":"info","ts":1644897063.4353817,"logger":"controller.lvmvolumegroup.vg-manager","msg":"does not match filter","reconciler group":"lvm.topolvm.io","reconciler kind":"LVMVolumeGroup","name":"vg1","namespace":"lvm-operator-system","Device.Name":"sdb","filter.Name":"notRemovable"}
{"level":"info","ts":1644897063.4425514,"logger":"controller.lvmvolumegroup.vg-manager","msg":"does not match filter","reconciler group":"lvm.topolvm.io","reconciler kind":"LVMVolumeGroup","name":"vg1","namespace":"lvm-operator-system","Device.Name":"sdc","filter.Name":"notRemovable"}
{"level":"info","ts":1644897063.449244,"logger":"controller.lvmvolumegroup.vg-manager","msg":"does not match filter","reconciler group":"lvm.topolvm.io","reconciler kind":"LVMVolumeGroup","name":"vg1","namespace":"lvm-operator-system","Device.Name":"sdd","filter.Name":"notRemovable"}```

**Comment by nbalacha** [comment]
We will take a look and get back to you.

**Comment by leelavg** [comment]
hi @deanpeterson 
- would you be able to provide the o/p of below command on the same machine on which you found the issue?
`lsblk -o NAME,ROTA,TYPE,SIZE,MODEL,VENDOR,RO,RM,STATE,KNAME,SERIAL,PARTLABEL,FSTYPE`

**Comment by garyedwards** [comment]
Hi would it be possible to get this patch committed to the `release-4.10` branch and pushed to OLM?

I followed the [official docs for SNO](https://access.redhat.com/documentation/en-us/red_hat_openshift_data_foundation/4.10/html/deploying_openshift_data_foundation_on_single_node_openshift_clusters/index) and I think this patch would solve the issue I am having with my SATA disks:

```
$ lsblk -o NAME,ROTA,TYPE,SIZE,MODEL,VENDOR,RO,RM,STATE,KNAME,SERIAL,PARTLABEL,FSTYPE
NAME        ROTA TYPE   SIZE MODEL                                    VENDOR   RO RM STATE   KNAME     SERIAL          PARTLABEL  FSTYPE
sda            1 disk   3.7T ST4000VN008-2DR1                         ATA       0  1 running sda       ZM40YCKP
[...]

PR #147, opened 2022-04-08
fix: do not filter out removable devices

leelavg:
- hot pluggable devices are represented as removable in lsblk o/p
- not all removable devices should be filtered out and this PR removes
  the existing filter
- as a side effect ROM is being picked up for VG ops and that is
  filtered based on Type of device

fixes: #109


leelavg:
- filtering out ROM device
``` json
{"level":"info","ts":1649414296.9406986,"logger":"controller.lvmvolumegroup.vg-manager","msg":"does not match filter","reconciler group":"lvm.topolvm.io","reconciler kind":"LVMVolumeGroup","name":"vg1","namespace":"lvm-operator-system","Device.Name":"sr0","filter.Name":"isUsableByDeviceType"}
```
- no regression
``` json
{"level":"info","ts":1649415065.6944656,"logger":"controller.lvmvolumegroup.vg-manager","msg":"does not match filter","reconciler group":"lvm.topolvm.io","reconciler kind":"LVMVolumeGroup","name":"vg1","namespace":"lvm-operator-system","Device.Name":"loop4","filter.Name":"isUsableByDeviceType"}
```

sp98 (approved):

pkg/vgmanager/filter.go
```diff
@@ -25,13 +25,12 @@ import (
...
 	// filter names:
 	notReadOnly           = "notReadOnly"
-	notRemovable          = "notRemovable"
 	notSuspended          = "notSuspended"
 	noBiosBootInPartLabel = "noBiosBootInPartLabel"
 	noFilesystemSignature = "noFilesystemSignature"
 	noBindMounts          = "noBindMounts"
 	noChildren            = "noChildren"
-	usableLoopDevice      = "usableLoopDevice"
+	isUsableByDeviceType  = "isUsableByDeviceType"
```
nbalacha, line 33:
```suggestion
	usableDeviceType  = "usableDeviceType"
```

nbalacha (changes requested):
If, in future, there is a requirement to be able to configure the VG to not use loop devices , how would this be implemented?

Would it better to have one filter per type and add them to the list of filters for a VG dynamically depending on the configuration?

leelavg:
> how would this be implemented?
- then it'll be a straightforward implementation, we use all the strings based on device types in single `case` of `switch` and will filter them out
- it basically boils out to filtering out devices based on device types currently loop is a special case

> Would it better to have one filter per type
- I feel adding filters per type will be too verbose and only increases filter count

nbalacha:
> > how would this be implemented?
> 
> * then it'll be a straightforward implementation, we use all the strings based on device types in single `case` of `switch` and will filter them out
> * it basically boils out to filtering out devices based on device types currently loop is a special case
> 
Can you elaborate on this seeing that the supported types would be dynamic?
> > Would it better to have one filter per type
> 
> * I feel adding filters per type will be too verbose and only increases filter count

nbalacha:
> > > how would this be implemented?
> > 
> > 
> > 
> > * then it'll be a straightforward implementation, we use all the strings based on device types in single `case` of `switch` and will filter them out
> > * it basically boils out to filtering out devices based on device types currently loop is a special case
> > 
> > > Would it better to have one filter per type
> > 
> > 
> > 
> > * I feel adding filters per type will be too verbose and only increases filter count
> 
> Agree with Leela here. I don't think we need multiple filters for each device class. The current switch case usage would suffice.

If, in future, we want to allow dynamic device type config, the switch case will not work. Say we want to use loopDevices for one cluster and not the other, we need a way to dynamically determine the filters.

nbalacha (approved):
Approving for now but we should relook at the filter approach.

PR #167, opened 2022-04-21
fix: add missing files for `make bundle`

Madhu-1:
After running `make bundle` locally few files got updated and new files also got added.

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

/cc @nbalacha

### DIFF
--- a/bundle/manifests/lvm-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/lvm-operator.clusterserviceversion.yaml
@@ -285,7 +285,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.name
-                image: quay.io/nibalach/lvm-operator:e2e
+                image: quay.io/ocs-dev/lvm-operator:latest
                 livenessProbe:
                   httpGet:
                     path: /healthz
@@ -310,7 +310,7 @@ spec:
                   allowPrivilegeEscalation: false
               - command:
                 - /metricsexporter
-                image: quay.io/nibalach/lvm-operator:e2e
+                image: quay.io/ocs-dev/lvm-operator:latest
                 name: metricsexporter
                 resources: {}
               securityContext:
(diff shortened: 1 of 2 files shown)


nbalacha (changes requested):
Some of these do not need to be updated. They are temporary files generated from templates.
However, [bundle/manifests/lvm-operator.clusterserviceversion.yaml](https://github.com/red-hat-storage/lvm-operator/pull/167/files#diff-f96bdaa8b095a2ecd1bfe9c839663e0352fe0ae96dda277b8de28fd1a7d96ab6) needs to be updated as the image is incorrect.

config/default/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
```
nbalacha, line 1: This was updated by the make bundle target as we use different ns for upstream and downstream.

bundle/manifests/lvm-operator.clusterserviceversion.yaml
```diff
@@ -285,7 +285,7 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.name
-                image: quay.io/nibalach/lvm-operator:e2e
+                image: quay.io/ocs-dev/lvm-operator:latest
```
nbalacha, line 288: This file should be committed.

config/default/manager_custom_env.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
```
nbalacha, line 1: This is a file generated by make update-mgr-env and does not need to be committed.

config/manager/kustomization.yaml
```diff
@@ -3,3 +3,10 @@ resources:
```
nbalacha, line 3: Not required to be committed.

Madhu-1: @nbalacha how about adding unwanted files to gitignore?

nbalacha: I don't think it is required. If it makes things less confusing, we can merge the temporary files as they will be overwritten anyway by the make targets.

Madhu-1: Yes as you mentioned it's less confusing and helps both contributors and maintainers to keep track of it.

nbalacha (approved):

PR #170, opened 2022-04-22
add suggested namespace to csv

yuvalk:
This is so that operator is always installed in openshift-storage
namespace.

see https://docs.openshift.com/container-platform/4.10/operators/operator_sdk/osdk-generating-csvs.html#osdk-suggested-namespace_osdk-generating-csvs


yuvalk:
this resolves #170

PR #175, opened 2022-04-27
[WIP]make operator running in any namespace

iamniting:


iamniting:
We tried to make it happen, But it does not seem to be possible for now because of the cluster role bindings we are using.

PR #186, opened 2022-05-06
fix: cleanup when pvcs are deleted

nbalacha:
The LVMCluster controller does not cleanup lvm resources
if there are PVCs provisioned by topolvm still existing when the
LVMCluster CR is deleted. This commit fixes an issue where the
LVMCluster finalizer is not removed even after the PVCs are deleted,
thus preventing the reseources from being cleaned up.

### DIFF
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -149,8 +149,6 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
 		}
 		if err != nil {
 			return ctrl.Result{Requeue: true, RequeueAfter: time.Minute * 1}, err
-		} else {
-			return reconcile.Result{}, nil
 		}
 	}
 


controllers/lvmcluster_controller.go
```diff
@@ -149,8 +149,6 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
 		}
 		if err != nil {
 			return ctrl.Result{Requeue: true, RequeueAfter: time.Minute * 1}, err
```
leelavg, line 151: - I'm sorry in failing to understand the scenario
- Would you pls expand a bit more?

t1: cr, finaliser, lv exist
t2: cr deleted, finaliser exist, lv failed to delete
t3: cr deleted, finaliser exist, lv deleted

this PR just removes finaliser after t3 in next reconcile?

sp98: confused as well. 
So now if the `r.processDelete` succeeds, the code flow will continue even after the `!instance.DeletionTimestamp.IsZero()` condition on line 137 ?

nbalacha: Quite right. I'm closing this PR.

PR #201, opened 2022-05-31
fix: creates the thinpool with the -Z option

nbalacha:
This commit updates the lvcreate command used to create
the thinpool with the -Z option to enable zeroing. This removes
the requirement that the thin_pool_zero option be enabled in the
lvm.conf.

### DIFF
--- a/pkg/vgmanager/vgmanager_controller.go
+++ b/pkg/vgmanager/vgmanager_controller.go
@@ -259,7 +259,7 @@ func (r *VGReconciler) addThinPoolToVG(vgName string, config *lvmv1alpha1.ThinPo
 		}
 	}
 
-	args := []string{"-l", fmt.Sprintf("%d%%FREE", config.SizePercent), "-c", DefaultChunkSize, "-T", fmt.Sprintf("%s/%s", vgName, config.Name)}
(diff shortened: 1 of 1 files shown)


pkg/vgmanager/vgmanager_controller.go
```diff
@@ -259,7 +259,7 @@ func (r *VGReconciler) addThinPoolToVG(vgName string, config *lvmv1alpha1.ThinPo
 		}
 	}
 
-	args := []string{"-l", fmt.Sprintf("%d%%FREE", config.SizePercent), "-c", DefaultChunkSize, "-T", fmt.Sprintf("%s/%s", vgName, config.Name)}
+	args := []string{"-l", fmt.Sprintf("%d%%FREE", config.SizePercent), "-c", DefaultChunkSize, "-Z", "y", "-T", fmt.Sprintf("%s/%s", vgName, config.Name)}
```
sp98, line 262: do we need `y`?  Default is yes

nbalacha: It is always better to be explicit.

sp98 (approved):

PR #217, opened 2022-06-23
Adds the recommended k8s labels to lvmo pods

nbalacha:


controllers/defaults.go
```diff
@@ -118,11 +117,24 @@ var (
...
+	// AppKubernetesManagedByLabel is the Kubernetes recommended managed-by label
+	AppKubernetesManagedByLabel = "app.kubernetes.io/managed-by"
+	// AppKubernetesComponentLabel is the Kubernetes recommended component label
+	AppKubernetesComponentLabel = "app.kubernetes.io/component"
+
+	DefaultLabelKey           = "app.lvm.openshift.io/name"
 	TopolvmControllerLabelVal = "topolvm-controller"
 	TopolvmNodeLabelVal       = "topolvm-node"
 	VGManagerLabelVal         = "vg-manager"
+	ManagedByLabelValue       = "lvm-operator"
```
sp98, line 135: some variables end with `val` and some with `value`.  We should make it consistent.

controllers/defaults.go
```diff
@@ -118,11 +117,24 @@ var (
 	// name of the lvm-operator container
 	LVMOperatorContainerName = "manager"
 
-	// default labels and values
-	DefaultLabelKey           = "app.lvm.openshift.io"
+	// labels and values
+
+	// AppKubernetesPartOfLabel is the Kubernetes recommended part-of label
+	AppKubernetesPartOfLabel = "app.kubernetes.io/part-of"
```
sp98, line 123: nit. Label keys could end with with `key`. `AppKubernetesPartOfLabelKey` or `PartOfLabelKey`

sp98 (changes requested):
Might want to update the selector labels in the [service](https://github.com/red-hat-storage/lvm-operator/blob/d79efebbed371bd9e444bf5fd6f2604e8daa1a6b/bundle/manifests/topolvm-node-metrics_v1_service.yaml#L15) as well

sp98 (approved):

PR #226, opened 2022-07-06
Use topolvm upstream

sp98:
Lvm operator was made to point to the downstream
topolvm branch because some dependent PRs were not
merged upstream. Now that all those PRs are merged
lvm operator can point back to upstream topolvm

Using 0.15.0 topolvm version


bundle/manifests/lvm-operator.clusterserviceversion.yaml
```diff
@@ -258,9 +258,7 @@ spec:
           - create
         serviceAccountName: controller-manager
       deployments:
-      - label:
-          app.kubernetes.io/name: lvm-operator
```
sp98, line 262: `make bundle` removes this label. Need to check again.

Yuggupta27 (changes requested):

sp98:
`make bundle` command removes the `label` from the CSV in operator sdk 1.14 version that I use. Nithya checked on 1.19 and it does not remove the label. 

So, its better to first upgrade the operator sdk version before merging this PR.

bundle/manifests/lvm-operator.clusterserviceversion.yaml
sp98: This is fixed after operator sdk version upgrade.

Makefile
```diff
@@ -110,7 +110,7 @@ test: manifests generate fmt vet envtest godeps-update ## Run tests.
 
 
 OPERATOR_NAMESPACE ?= openshift-storage
-TOPOLVM_CSI_IMAGE ?= quay.io/ocs-dev/topolvm:latest
+TOPOLVM_CSI_IMAGE ?= quay.io/topolvm/topolvm:0.15.0
```
nbalacha, line 113: This image does not exist upstream which is why the CI is failing.
Lets stick to quay.io/ocs-dev/topolvm:latest

go.mod
```diff
@@ -113,5 +113,3 @@ require (
 	sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 )
-
-replace github.com/topolvm/topolvm => github.com/red-hat-storage/topolvm v0.11.1-0.20220429122749-ebdea64ffc67
```
nbalacha, line 117: I think this is the only change we really require.

PR #229, opened 2022-07-18
feat: make API and controller changes to support manual disk selection

iamniting:
Allow user to select the block devices while VG creation in the LVMCluster CR.


api/v1alpha1/lvmcluster_types.go
```diff
@@ -90,6 +90,10 @@ type DeviceSelector struct {
 	// MinSize is the minimum size of the device which needs to be included. Defaults to `1Gi` if empty
 	// +optional
 	// MinSize *resource.Quantity `json:"minSize,omitempty"`
+
+	// Paths are the complete location of block devices, which needs to be included while creating a VG.
+	// +optional
+	Paths []string `json:"paths,omitempty"`
```
sp98, line 96: `DevicePaths` should be a better name than `Paths`, IMO.

api/v1alpha1/lvmcluster_types.go
```diff
@@ -90,6 +90,10 @@ type DeviceSelector struct {
 	// MinSize is the minimum size of the device which needs to be included. Defaults to `1Gi` if empty
 	// +optional
 	// MinSize *resource.Quantity `json:"minSize,omitempty"`
+
+	// Paths are the complete location of block devices, which needs to be included while creating a VG.
```
sp98, line 94: // A list of device paths which would be chosen creating Volume Group.

api/v1alpha1/lvmvolumegroup_types.go
```diff
@@ -46,6 +46,8 @@ type LVMVolumeGroupSpec struct {
 
 // LVMVolumeGroupStatus defines the observed state of LVMVolumeGroup
 type LVMVolumeGroupStatus struct {
+	// LVMVGStatus contains the status of the VG
+	LVMVGStatus VGStatus `json:"lvmVgStatus,omitempty"`
```
sp98, line 50: Didn't understand why this change is part of the manual disk selection api changes!

sp98 (changes requested):

api/v1alpha1/lvmcluster_types.go
sp98: The comment should also discourage from using device names. Also, it would be good idea to include examples of the device paths ( like `/dev/disk/by-path/... ` etc) in the comment.

api/v1alpha1/lvmcluster_types.go
sp98: I see now that's this field is already inside `DeviceSelector`.  So `Paths` is also fine.

api/v1alpha1/lvmvolumegroup_types.go
iamniting: currently, we do not set any status on the volume group CR. This means we can not communicate the failure to the user easily. So we should set the status and communicate the failures or success via the status

sp98: We do that in `LVMVolumeGroupNodeStatusSpec` of a different CR.

iamniting: But that is only on a success right not on a failure

api/v1alpha1/lvmcluster_types.go
iamniting: keeping it as paths for now

api/v1alpha1/lvmcluster_types.go
iamniting: changed the description and added the example as well

api/v1alpha1/lvmvolumegroup_types.go
iamniting: I removed this as it will create conflict while running vgmanager on the multiple nodes.

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -254,6 +254,10 @@ spec:
                           node:
                             description: Node is the name of the node
                             type: string
+                          reason:
```
sp98, line 257: This looks like something that can be part of the `make bundle` commit.

controllers/lvmcluster_controller.go
```diff
@@ -208,6 +215,46 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
 	return ctrl.Result{}, nil
 }
 
+func (r *LVMClusterReconciler) verifyLvmClusterSpec(ctx context.Context, instance *lvmv1alpha1.LVMCluster) error {
+
+	// make sure no device overlap with another VGs
```
sp98, line 220: what if the user decides to give the device names instead of device paths?
I know we encourage them to use paths, but still if they decide to use device names instead, then this validation will always fail, if I'm not wrong.

controllers/lvmcluster_controller.go
```diff
@@ -208,6 +215,46 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
 	return ctrl.Result{}, nil
 }
 
+func (r *LVMClusterReconciler) verifyLvmClusterSpec(ctx context.Context, instance *lvmv1alpha1.LVMCluster) error {
```
sp98, line 218: method name can be more specific about what we are trying to verify instead of `verifyLvmClusterSpec`

sp98 (changes requested):

api/v1alpha1/lvmcluster_types.go
```diff
@@ -108,6 +114,11 @@ type LVMClusterStatus struct {
 	// Ready describes if the LVMCluster is ready.
 	// +optional
 	Ready bool `json:"ready,omitempty"`
+
+	// Reason describes the reason behind any success or failure.
+	// +optional
+	Reason string `json:"reason,omitempty"`
```
sp98, line 120: Do we add `reason` for error specific to this PR only?
If that's the case, then we need to add reasons for other failures as well in a separate PR.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,12 +399,86 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
 }
 
 // filterMatchingDevices returns unmatched and matched blockdevices
-// TODO: Implement this
-func filterMatchingDevices(blockDevices []internal.BlockDevice, volumeGroup *lvmv1alpha1.LVMVolumeGroup) ([]internal.BlockDevice, []internal.BlockDevice, error) {
-	// currently just match all devices
+func (r *VGReconciler) filterMatchingDevices(blockDevices []internal.BlockDevice, volumeGroup *lvmv1alpha1.LVMVolumeGroup) ([]internal.BlockDevice, []internal.BlockDevice, error) {
```
sp98, line 400: Unit test should be added for this function.

sp98, line 400: do you plan to the use the `unmatched disks` return value? If not, then we can avoid returning that.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,12 +397,85 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+				} else {
+					err = fmt.Errorf("can not find disk path %s disk name %s in the available block devices", path, diskName)
+					return []internal.BlockDevice{}, []internal.BlockDevice{}, err
+				}
+			} else if filepath.Dir(path) == internal.DiskByNamePrefix {
+				// handle disk by names here such as /dev/nvme0n1
+				if ok {
+					filteredBlockDevices = append(filteredBlockDevices, blockDevice)
+				} else {
+					err := fmt.Errorf("can not find disk name %s in the available block devices", path)
```
sp98, line 439:
```suggestion
					err := fmt.Errorf("can not find disk name %s in the available block devices", diskName)
```

sp98 (changes requested):

controllers/lvmcluster_controller.go
iamniting: Even in that case, it should work, as the device names will only be the same. I am just loading the devices into a map; if the key already exists, we are erroring out.

api/v1alpha1/lvmcluster_types.go
```diff
@@ -108,6 +114,11 @@ type LVMClusterStatus struct {
 	// Ready describes if the LVMCluster is ready.
 	// +optional
 	Ready bool `json:"ready,omitempty"`
+
+	// Reason describes the reason behind any success or failure.
```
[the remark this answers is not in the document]
iamniting: we can also have a reason behind the success which will always be the exact string. I think success reason should not matter much, as much as failure.

controllers/lvmcluster_controller.go
```diff
@@ -208,6 +215,46 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
+					} else {
+						err = fmt.Errorf("Error: device path %s is specified at multiple places in deviceClass %s", path, val)
+					}
+					r.Log.Error(err, "failed to verify device paths")
+
+					instance.Status.Ready = false
+					instance.Status.Reason = err.Error()
+
+					// Apply status changes
+					err = r.Client.Status().Update(ctx, instance)
```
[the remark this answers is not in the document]
iamniting: This verification function is called precisely after getting the instance and we won't get to an actual update call as we will be returning from here only, so we need to do an update here itself only and only if there is a failure.

api/v1alpha1/lvmcluster_types.go
iamniting: I am updating the reason wherever required.

pkg/vgmanager/vgmanager_controller.go
iamniting: no plans we can remove it

pkg/vgmanager/vgmanager_controller.go
iamniting: both will be the same only, So updating to path should make any differences

nbalacha (changes requested):

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -373,7 +372,11 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
 
 	args := []string{vgName}
 	for _, device := range devices {
-		args = append(args, fmt.Sprintf("/dev/%s", device.KName))
+		if device.DiskByPath != "" {
+			args = append(args, device.DiskByPath)
+		} else {
+			args = append(args, device.KName)
```
sp98, line 378: don't think it will work if complete device name with path (for example: `dev/sdb`)  is not provided in the args

pkg/internal/block_device.go
```diff
@@ -39,17 +43,20 @@ type BlockDevice struct {
...
+	DiskByPath string
 }
 
 // ListBlockDevices using the lsblk command
 func ListBlockDevices(exec Executor) ([]BlockDevice, error) {
 	// var output bytes.Buffer
 	var blockDeviceMap map[string][]BlockDevice
 	columns := "NAME,ROTA,TYPE,SIZE,MODEL,VENDOR,RO,STATE,KNAME,SERIAL,PARTLABEL,FSTYPE"
-	args := []string{"--json", "-o", columns}
+	args := []string{"--json", "--paths", "-o", columns}
```
sp98, line 59: why is `--paths` argument needed?

pkg/vgmanager/vgmanager_controller.go
iamniting: It should work as we are getting the full path now.

pkg/internal/block_device.go
iamniting: using --paths will give us the full path and make our job easy while comparing the paths and while adding paths into some commands we don't have to add `/dev/` anymore.

api/v1alpha1/lvmcluster_types.go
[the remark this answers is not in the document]
iamniting: added conditions

controllers/lvmcluster_controller.go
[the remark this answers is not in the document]
iamniting: checking node selectors as well now

pkg/vgmanager/vgmanager_controller.go
sp98: Still see this return value in the code.

sp98: comment still valid.

sp98 (commented):
some comments are not addressed

pkg/vgmanager/vgmanager_controller.go
iamniting: I will add them later once we agree with the actual code changes to minimize the changes in unit tests.

iamniting: missed it earlier, made changes now

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -47,6 +47,15 @@ spec:
                         deviceSelector:
                           description: DeviceSelector is a set of rules that should
                             match for a device to be included in the LVMCluster
+                          properties:
+                            paths:
+                              description: A list of device paths which would be chosen
+                                for creating Volume Group. for example "/dev/disk/by-path/pci-0000:04:00.0-nvme-1"
```
sp98, line 53:
```suggestion
                                for creating Volume Group. For example "/dev/disk/by-path/pci-0000:04:00.0-nvme-1"
```

api/v1alpha1/lvmcluster_types.go
```diff
@@ -17,13 +17,19 @@ limitations under the License.
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	// ConditionLVMClusterCRvalid communicates if LVMCluster CR is invalid
+	ConditionLVMClusterCRvalid conditionsv1.ConditionType = "LVMClusterCRvalid"
```
sp98, line 30: can we remove the `CR` from this name?  I think LVMCluster and LVMClusterCR both should imply the same thing.

controllers/conditions.go
```diff
@@ -0,0 +1,35 @@
...
+package controllers
+
+import (
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
+)
+
+func SetLVMClusterCRvalidCondition(conditions *[]conditionsv1.Condition,
```
sp98, line 26: does it need to be an exported function?

controllers/lvm_volumegroup.go
```diff
@@ -43,8 +43,14 @@ func (c lvmVG) ensureCreated(r *LVMClusterReconciler, ctx context.Context, lvmCl
 	lvmVolumeGroups := c.getLvmVolumeGroups(r, lvmCluster)
 
 	for _, volumeGroup := range lvmVolumeGroups {
-		result, err := cutil.CreateOrUpdate(ctx, r.Client, volumeGroup, func() error {
-			// no need to mutate any field
+		existingvolumeGroup := &lvmv1alpha1.LVMVolumeGroup{
```
sp98, line 46:
```suggestion
		existingVolumeGroup := &lvmv1alpha1.LVMVolumeGroup{
```

controllers/lvmcluster_controller.go
sp98: This comments still holds. The function only validates the device classes in the Spec. The function name should reflect that.

sp98 (changes requested):

api/v1alpha1/lvmcluster_types.go
nbalacha: Agreed.

controllers/lvmcluster_controller.go
```diff
@@ -103,6 +103,13 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
+
+	err = r.verifyLvmClusterSpec(ctx, lvmCluster)
+	if err != nil {
+		r.Log.Error(err, "failed to verify LVM cluster spec")
```
nbalacha, line 109:
```suggestion
		r.Log.Error(err, "failed to verify LVMCluster")
```

nbalacha: Or "invalid LVMCluster Spec."

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
iamniting: done

controllers/conditions.go
iamniting: not really, unexported it.

api/v1alpha1/lvmcluster_types.go
iamniting: done

controllers/lvm_volumegroup.go
iamniting: done

controllers/lvmcluster_controller.go
iamniting: done

controllers/lvmcluster_controller.go
iamniting: added a new func for this logic and that will be called from here

go.mod
```diff
@@ -11,8 +12,8 @@ require (
 	github.com/onsi/gomega v1.19.0
 	github.com/openshift/api v0.0.0-20211028023115-7224b732cc14
 	github.com/openshift/client-go v0.0.0-20210831095141-e19a065e79f7
+	github.com/openshift/custom-resource-status v1.1.2
 	github.com/operator-framework/api v0.14.0
-	github.com/operator-framework/operator-lifecycle-manager v0.20.0
```
nbalacha, line 15: We use OLM for e2e tests. Will removing this cause issues?

sp98 (approved):

go.mod
iamniting: I ran go mod tidy and it removed it automatically, which means we are not using it anymore.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,10 +397,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+				} else {
+					err = fmt.Errorf("can not find disk path %s disk name %s in the available block devices", path, diskName)
+					return []internal.BlockDevice{}, err
+				}
+			} else if filepath.Dir(path) == internal.DiskByNamePrefix {
+				// handle disk by names here such as /dev/nvme0n1
+				if ok {
+					filteredBlockDevices = append(filteredBlockDevices, blockDevice)
+				} else {
+					err := fmt.Errorf("can not find disk name %s in the available block devices", path)
```
nbalacha, line 439:
```suggestion
					err := fmt.Errorf("can not find device name %s in the available block devices", path)
```

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,10 +397,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+
+			blockDevice, ok := hasExactDisk(blockDevices, diskName)
+
+			if filepath.Dir(path) == internal.DiskByPathPrefix {
+				// handle disk by path here such as /dev/disk/by-path/pci-0000:87:00.0-nvme-1
+				if ok {
+					blockDevice.DiskByPath = path
+					filteredBlockDevices = append(filteredBlockDevices, blockDevice)
+				} else {
+					err = fmt.Errorf("can not find disk path %s disk name %s in the available block devices", path, diskName)
```
nbalacha, line 431:
```suggestion
					err = fmt.Errorf("can not find device path %s, device name %s in the available block devices", path, diskName)
```

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -488,15 +565,50 @@ func (r *VGReconciler) generateVolumeGroupNodeStatus() (*lvmv1alpha1.LVMVolumeGr
 	//lvmvgstatus := vgNodeStatus.Spec.LVMVGStatus
 	var statusarr []lvmv1alpha1.VGStatus
 
+	var isVgExist bool
+
 	for _, vg := range vgs {
+
```
nbalacha, line 571: Nit: blank line

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -488,15 +565,50 @@ func (r *VGReconciler) generateVolumeGroupNodeStatus() (*lvmv1alpha1.LVMVolumeGr
...
+		// Add pvs as per volumeGroup CR if path is given add path else add name
+		diskPattern := internal.DiskByNamePrefix
+
```
nbalacha, line 574: There could be vgs that are not created by LVMO.  Please skip the processing in such a case.

nbalacha (changes requested):

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,10 +397,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+				return []internal.BlockDevice{}, err
+			}
+
+			isAlreadyExist := isDeviceAlreadyAddedtoVG(vgs, diskName, volumeGroup)
+			if isAlreadyExist {
+				continue
+			}
+
+			blockDevice, ok := hasExactDisk(blockDevices, diskName)
+
```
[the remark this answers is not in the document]
iamniting: Actually, I am doing that for a reason. I am creating 2 different errors based on the path type. if I follow this approach I will end up having one error only

pkg/vgmanager/vgmanager_controller.go
iamniting: I would suggest let's take it up in the next PR. So that QE will have something to test ASAP.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -181,9 +178,11 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
 	// create/extend VG and update lvmd config
 	err = r.addDevicesToVG(volumeGroup.Name, matchingDevices)
 	if err != nil {
+		status.Reason = err.Error()
```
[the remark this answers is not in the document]
iamniting: Yes, it should be enough, We are already sending errors via appending a string or an error is generated by us in somecases.

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -148,18 +141,22 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
...
-		volGrpHostInfo, err := GetVolumeGroup(r.executor, volumeGroup.Name)
-		if err != nil {
-			r.Log.Error(err, "failed to get volume group from the host", "VGName", volumeGroup.Name)
-		} else {
-			status.Devices = volGrpHostInfo.PVs
+
+	//Get the block devices that can be used for this volumegroup
+	matchingDevices, delayedDevices, err := r.getMatchingDevicesForVG(volumeGroup)
+	if err != nil {
+		status.Reason = err.Error()
```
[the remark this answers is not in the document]
iamniting: same here

pkg/vgmanager/vgmanager_controller.go
nbalacha: It is quite a small change - if !ok { continue }.

iamniting: done

pkg/vgmanager/vgmanager_controller.go
iamniting: done

pkg/vgmanager/vgmanager_controller.go
iamniting: done

pkg/vgmanager/vgmanager_controller.go
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -488,15 +565,50 @@ func (r *VGReconciler) generateVolumeGroupNodeStatus() (*lvmv1alpha1.LVMVolumeGr
 	//lvmvgstatus := vgNodeStatus.Spec.LVMVGStatus
 	var statusarr []lvmv1alpha1.VGStatus
 
+	var isVgExist bool
```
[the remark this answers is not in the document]
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,10 +397,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+func hasExactDisk(blockDevices []internal.BlockDevice, deviceName string) (internal.BlockDevice, bool) {
+	for _, blockDevice := range blockDevices {
+		if blockDevice.KName == deviceName {
+			return blockDevice, true
+		}
+	}
+	return internal.BlockDevice{}, false
+}
+
+func isDeviceAlreadyAddedtoVG(vgs []VolumeGroup, diskName string, volumeGroup *lvmv1alpha1.LVMVolumeGroup) bool {
```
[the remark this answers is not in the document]
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -394,10 +397,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+	var filteredBlockDevices []internal.BlockDevice
+
+	if volumeGroup.Spec.DeviceSelector != nil && len(volumeGroup.Spec.DeviceSelector.Paths) > 0 {
+		vgs, err := ListVolumeGroups(r.executor)
+		if err != nil {
+			return []internal.BlockDevice{}, fmt.Errorf("failed to list volume groups. %v", err)
+		}
+
+		for _, path := range volumeGroup.Spec.DeviceSelector.Paths {
+
```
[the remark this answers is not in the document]
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -148,18 +141,22 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
...
+	matchingDevices, delayedDevices, err := r.getMatchingDevicesForVG(volumeGroup)
+	if err != nil {
+		status.Reason = err.Error()
+		if statuserr := r.updateStatus(ctx, status); statuserr != nil {
+			r.Log.Error(statuserr, "failed to update status", "name", volumeGroup.Name)
+			return reconcileAgain, nil
 		}
+
+		// Failed to get devices for this vg. Reconcile again.
+		r.Log.Error(err, "failed to get block devices for volumegroup", "VGName", volumeGroup.Name)
```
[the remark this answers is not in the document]
iamniting: done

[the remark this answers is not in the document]
iamniting: done

pkg/internal/block_device.go
```diff
@@ -64,6 +71,29 @@ func ListBlockDevices(exec Executor) ([]BlockDevice, error) {
...
+	paths, err := filepath.Glob(diskByPathDir)
+	if err != nil {
+		return nil, fmt.Errorf("could not list files in %q: %w", diskByPathDir, err)
+	}
+
+	for _, path := range paths {
+		diskName, err := filepath.EvalSymlinks(path)
+		if err != nil {
+			return nil, fmt.Errorf("could not eval symLink %q:%w", path, err)
+
```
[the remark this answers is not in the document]
iamniting: done

pkg/internal/block_device.go
```diff
@@ -64,6 +71,29 @@ func ListBlockDevices(exec Executor) ([]BlockDevice, error) {
...
 
+func ListDiskByPath(exec Executor) (map[string]string, error) {
+
+	devices := make(map[string]string)
+
+	diskByPathDir := filepath.Join(DiskByPathPrefix, "/*")
+
+	paths, err := filepath.Glob(diskByPathDir)
+	if err != nil {
+		return nil, fmt.Errorf("could not list files in %q: %w", diskByPathDir, err)
```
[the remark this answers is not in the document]
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -541,3 +663,21 @@ func (r *VGReconciler) updateStatus(ctx context.Context) error {
...
 }
+
+func (r *VGReconciler) getAllLvmVolumeGroups(ctx context.Context) (map[string]*lvmv1alpha1.LVMVolumeGroup, error) {
+
+	lvmVolumeGroupsMap := make(map[string]*lvmv1alpha1.LVMVolumeGroup)
+
+	lvmVolumeGroups := &lvmv1alpha1.LVMVolumeGroupList{}
+	err := r.Client.List(ctx, lvmVolumeGroups, &client.ListOptions{Namespace: r.Namespace})
+	if err != nil {
+		r.Log.Error(err, "failed to list VolumeGroups")
```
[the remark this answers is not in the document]
iamniting: done

controllers/lvmcluster_controller.go
```diff
@@ -220,6 +227,71 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
+		    "nodeSelector2": {
+		        "/dev/sda": "vg1",
+		        "/dev/sdb": "vg1"
+		    }
+		}
+	*/
+	devices := make(map[string]map[string]string)
+
+	for _, deviceClass := range instance.Spec.Storage.DeviceClasses {
+		if deviceClass.DeviceSelector != nil {
```
mulbc, line 265: what if we (in the future) have two deviceClasses - one taking up all the disks and one with manual disk selection?

controllers/lvmcluster_controller.go
```diff
@@ -220,6 +227,71 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
+					var err error
+					if val != deviceClass.Name {
+						err = fmt.Errorf("Error: device path %s overlaps in two different deviceClasss %s and %s", path, val, deviceClass.Name)
+					} else {
+						err = fmt.Errorf("Error: device path %s is specified at multiple places in deviceClass %s", path, val)
+					}
+					r.Log.Error(err, "failed to verify device paths")
+
+					setLVMClusterCRvalidCondition(&instance.Status.Conditions, corev1.ConditionFalse, "CRInvalid", err.Error())
+					return err
```
mulbc, line 278: shouldn't we use `setLVMClusterCRvalidCondition` when there's a verify issue?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -269,6 +274,45 @@ func (r *VGReconciler) addThinPoolToVG(vgName string, config *lvmv1alpha1.ThinPo
...
+		return fmt.Errorf("failed to parse lvSize. %v", err)
+	}
+
+	vgSize, err := strconv.ParseFloat(vg.VgSize[:len(vg.VgSize)-1], 64)
+	if err != nil {
+		return fmt.Errorf("failed to parse vgSize. %v", err)
+	}
+
+	// return if thinPoolSize is as expected
+	if config.SizePercent == int((thinPoolSize/vgSize)*100) {
```
mulbc, line 298: this should be <= instead of ==
we don't want to run lvextend if the thinpoolsize is larger than config.SizePercent (I hope we never get there though ;) )

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -393,11 +441,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+			} else if filepath.Dir(path) == internal.DiskByNamePrefix {
+				// handle disk by names here such as /dev/nvme0n1
+				if ok {
+					filteredBlockDevices = append(filteredBlockDevices, blockDevice)
+				} else {
+					err := fmt.Errorf("can not find device name %s in the available block devices", path)
+					return []internal.BlockDevice{}, err
+				}
+			} else {
+				err = fmt.Errorf("unsupported disk path format %s", path)
```
mulbc, line 487: please add to the error message that only `/dev/disk/by-path` and `/dev/` links are currently supported

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -393,11 +441,83 @@ func (r *VGReconciler) addDevicesToVG(vgName string, devices []internal.BlockDev
...
+			}
+
+			isAlreadyExist := isDeviceAlreadyPartOfVG(vgs, diskName, volumeGroup)
+			if isAlreadyExist {
+				continue
+			}
+
+			blockDevice, ok := hasExactDisk(blockDevices, diskName)
+
+			if filepath.Dir(path) == internal.DiskByPathPrefix {
```
mulbc, line 469: why can't we use this flow for all `/dev/disk/by-*` options?

controllers/lvmcluster_controller.go
iamniting: We are setting the condition in both cases either in failure or in success.

Can you pls elaborate on what you mean?

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -488,15 +609,52 @@ func (r *VGReconciler) generateVolumeGroupNodeStatus() (*lvmv1alpha1.LVMVolumeGr
...
+
+		deviceSelector := lvmVolumeGroup.Spec.DeviceSelector
+		if deviceSelector != nil && len(deviceSelector.Paths) > 0 {
+			if filepath.Dir(deviceSelector.Paths[0]) == internal.DiskByPathPrefix {
+				diskPattern = internal.DiskByPathPrefix
+			}
+		}
+
+		devices := []string{}
+		if diskPattern == internal.DiskByPathPrefix {
```
mulbc, line 631: idea: why not always use the disk path in the status?!

controllers/lvmcluster_controller.go
mulbc: I need more coffee :D sorry

controllers/lvmcluster_controller.go
iamniting: we should not support such cases IMO. This code will just check the manual disk paths only as other device class wont have any list of devices in the spec.

pkg/vgmanager/vgmanager_controller.go
iamniting: done

pkg/vgmanager/vgmanager_controller.go
iamniting: done

pkg/vgmanager/vgmanager_controller.go
iamniting: We can, We need more conditions to do that. Also as of now, all other patterns other than diskbypath will go into the else condition starting with dev.

pkg/vgmanager/vgmanager_controller.go
iamniting: I had this discussion earlier with @nbalacha and we came to the conclusion that we will show the paths as given in the spec. It will be easier for a user to compare them from the spec.

controllers/lvmcluster_controller.go
```diff
@@ -220,6 +227,71 @@ func (r *LVMClusterReconciler) reconcile(ctx context.Context, instance *lvmv1alp
...
+				if devices[nodeSelector] == nil {
+					devices[nodeSelector] = make(map[string]string)
+				}
+
+				devices[nodeSelector][path] = deviceClass.Name
+			}
+		}
+	}
+
+	setLVMClusterCRvalidCondition(&instance.Status.Conditions, corev1.ConditionTrue, "CRValid", "CR is validated successfully")
```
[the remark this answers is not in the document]
iamniting: done

pkg/vgmanager/vgmanager_controller.go
```diff
@@ -148,18 +141,23 @@ func (r *VGReconciler) reconcile(ctx context.Context, req ctrl.Request, volumeGr
 	}
 
 	status := &lvmv1alpha1.VGStatus{
-		Name:   req.Name,
-		Status: lvmv1alpha1.VGStatusReady,
```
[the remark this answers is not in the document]
iamniting: I was not setting ready, but I was setting the degraded and failed. Fixed it.

nbalacha (approved):

PR #230, opened 2022-07-25
fix: use downloaded operator-sdk

nbalacha:
This commit updates the Makefile to use the downloaded
operator-sdk so as to use a fixed version on every system.
It also updates the operator-sdk version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.0.1
 
+OPERATOR_SDK_VERSION ?= 1.21.0
+
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
 # To re-generate a bundle for other specific channels without changing the standard setup, you can:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -268,20 +270,12 @@ endif
 
 .PHONY: operator-sdk
 OPERATOR_SDK = ./bin/operator-sdk
-operator-sdk: ## Download operator-sdk locally if necessary.
-ifeq (,$(wildcard $(OPERATOR_SDK)))
```
leelavg, line 272: - better have this one atleast or-else sdk is downloaded everytime the target is called
- the downside will be a mismatch version, however that's better than downloading everytime I believe.

nbalacha: The intention is to use the same version everytime as we are seeing different behaviour with different versions.

leelavg: yes, if the frequency of calling this target as a sub-target is less I'd also vote for this PR.

Makefile
```diff
@@ -5,6 +5,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 0.0.1
 
+OPERATOR_SDK_VERSION ?= 1.21.0
```
sp98, line 8: why not the latest version `1.22.2`?

nbalacha: There are breaking changes which I would like ti understand before upgrading to this version.

sp98 (approved):

PR #232, opened 2022-08-01
feat: validate nodeSelector of the lvmCluster

sp98:
Return error if the LvmCluster deviceClass nodeSelector filter does not match any node.

Signed-off-by: Santosh Pillai <sapillai@redhat.com>

### DIFF
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -79,6 +79,7 @@ type LVMClusterReconciler struct {
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;create;update;delete
 //+kubebuilder:rbac:groups=topolvm.cybozu.com,resources=logicalvolumes,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -32,6 +32,7 @@ import (
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
 	topolvmv1 "github.com/topolvm/topolvm/api/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	v1helper "k8s.io/component-helpers/scheduling/corev1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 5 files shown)


controllers/suite_test.go
```diff
@@ -111,6 +113,21 @@ var _ = BeforeSuite(func() {
 	testNamespace.Name = testLvmClusterNamespace
 	Expect(k8sClient.Create(ctx, testNamespace)).Should(Succeed())
 
+	testNode := &corev1.Node{
+
```
nbalacha, line 117: Nit. Please remove the blank line.

controllers/suite_test.go
```diff
@@ -111,6 +113,21 @@ var _ = BeforeSuite(func() {
...
+			Kind: "Node",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "node1",
+			Labels: map[string]string{
+				hostNameLabel: "node1",
+			},
+		},
+	}
+
```
nbalacha, line 128: Nit. Please remove the blank line.

sp98: fixed

controllers/lvmcluster_controller.go
```diff
@@ -79,6 +79,7 @@ type LVMClusterReconciler struct {
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;create;update;delete
 //+kubebuilder:rbac:groups=topolvm.cybozu.com,resources=logicalvolumes,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
```
iamniting, line 82: Is not get and list would be enough?

controllers/lvmcluster_controller.go
```diff
@@ -115,6 +117,40 @@ func (r *LVMClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	nodes := &corev1.NodeList{}
+	err = r.Client.List(ctx, nodes)
+	if err != nil {
+		r.Log.Error(err, "failed to list nodes")
+		return ctrl.Result{}, err
+	}
+
+	if len(nodes.Items) == 0 {
+		r.Log.Error(err, "no nodes are available")
+		return ctrl.Result{}, err
```
iamniting, line 130: err will be nil here, Do you really want to return nil here?

controllers/lvmcluster_controller_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	. "github.com/onsi/gomega"
 	lvmv1alpha1 "github.com/red-hat-storage/lvm-operator/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
```
iamniting, line 25: i would suggest rename it to corev1

iamniting (changes requested):

controllers/lvmcluster_controller.go
sp98: I though it would  but I was getting issues around `watch`. Don't have those logs right now. Adding `watch` as well fixed it.

controllers/lvmcluster_controller_test.go
sp98: done.

iamniting (approved):
/lgtm

nbalacha:
I'm not convinced that we really need this check. No daemonset pods will come up so I think it should be ok not to have this. A possible approach would be to update the status using the resource status fields (so if no daemonset pods are running, use that that status).
If we do want a node selector check, is there a way to pass the selectors to the List call to only return those nodes that match the filter?

sp98:
> If we do want a node selector check, is there a way to pass the selectors to the List call to only return those nodes that match the filter?

I checked a bit but didn't a way to do this.

sp98:
> I'm not convinced that we really need this check. No daemonset pods will come up so I think it should be ok not to have this. A possible approach would be to update the status using the resource status fields (so if no daemonset pods are running, use that that status).

We create more than one daemonsets. We might have to check all of them to ascertain that no daemonset pods are running.

nbalacha:
> > I'm not convinced that we really need this check. No daemonset pods will come up so I think it should be ok not to have this. A possible approach would be to update the status using the resource status fields (so if no daemonset pods are running, use that that status).
> 
> We create more than one daemonsets. We might have to check all of them to ascertain that no daemonset pods are running.

That is correct but checking nodes on a large cluster is not going to be easy or fast.

PR #235, opened 2022-08-03
test: added test for validating data in snapshot

riya-singhal31:
this test adds the data to volume, and checks whether the data is
successfully restored in snapshot or not.

Signed-off-by: riya-singhal31 <rsinghal@redhat.com>


e2e/config.go
```diff
@@ -37,8 +39,28 @@ var (
 	lvmOperatorUninstall bool
 	scheme               = runtime.NewScheme()
 	crClient             crclient.Client
+	ClientSet            clientset.Interface
+	parameterCodec       runtime.ParameterCodec
 )
 
+type Interface interface {
```
sp98, line 46: Changes in this file doesn't look right.  Looks right the code from the library!

e2e/config.go
```diff
@@ -37,8 +40,19 @@ var (
 	lvmOperatorUninstall bool
 	scheme               = runtime.NewScheme()
 	crClient             crclient.Client
+	parameterCodec       runtime.ParameterCodec
+	clientset            *kubernetes.Clientset
 )
 
+type Clientset struct {
+	*corev1.CoreV1Client
+}
```
nbalacha, line 49: I don't think this is required.

e2e/testdata.go
```diff
@@ -96,3 +111,120 @@ func getSamplePvc(size, name string, volumemode k8sv1.PersistentVolumeMode, stor
...
+}
+
+// ExecWithOptions executes a command in the specified container,
+// returning stdout, stderr and error. `options` allowed for
+// additional parameters to be passed.
+func ExecWithOptions(options ExecOptions) (string, string, error) {
+	if !options.Quiet {
+		fmt.Printf("ExecWithOptions %+v", options)
+	}
+	config, err := getKubeconfig("")
```
nbalacha, line 149: This is already done in the init function. Why is it require again?

e2e/config.go
```diff
@@ -37,8 +40,19 @@ var (
...
+
+// GetParameterCodec is the function used to retrieve the parameterCodec
+func GetParameterCodec() runtime.ParameterCodec {
+	return parameterCodec
```
nbalacha, line 53: where has the parameterCodec been initialised?

nbalacha (changes requested):

e2e/config.go
riya-singhal31: Updated.

e2e/config.go
riya-singhal31: Yes, updated.

e2e/config.go
riya-singhal31: Corrected.

e2e/testdata.go
```diff
@@ -37,6 +45,12 @@ func getSamplePod(name, pvcName string) *k8sv1.Pod {
 							ContainerPort: 80,
 						},
 					},
+					VolumeMounts: []k8sv1.VolumeMount{
```
nbalacha, line 48: This will only work for VolumeMode == Filesystem. For Block, it needs to be:

```
      volumeDevices:
        - name: data
          devicePath: /dev/xvda
```
See https://kubernetes.io/docs/concepts/storage/persistent-volumes/#pod-specification-adding-raw-block-device-path-in-container

nbalacha:
Please resolve the conflicts listed.

e2e/testdata.go
```diff
@@ -96,3 +121,133 @@ func getSamplePvc(size, name string, volumemode k8sv1.PersistentVolumeMode, stor
 	}
 	return pvc
 }
+
+// ExecOptions passed to ExecWithOptions
+type ExecOptions struct {
+	Command       []string
```
nbalacha, line 127: The exec commands defined here are already available in k8s.io/kubernetes/test/e2e/framework. Please use the functions from that package instead.

nbalacha (changes requested):
Please address the following:

- Use the exec* functions from k8s.io/kubernetes/test/e2e/framework
- Handle the conflicts.

e2e/testdata.go
```diff
@@ -0,0 +1,143 @@
...
+func writeDataInPod(pod *k8sv1.Pod, mode string) error {
+	gomega.Expect(pod.Spec.Containers).NotTo(gomega.BeEmpty())
+	var filePath string
+	if mode == "file" {
+		filePath = pod.Spec.Containers[0].VolumeMounts[0].MountPath + "/test"
+	} else {
+		filePath = pod.Spec.Containers[0].VolumeDevices[0].DevicePath
+	}
+
+	command := fmt.Sprintf("dd if=/dev/zero of=%s bs=1M count=10 status=none", filePath)
```
nbalacha, line 122: Please write something else to the file and check the contents in validateDataInPod.

PR #237, opened 2022-08-16
fix: create e2e resources from yaml files

nbalacha:
Modified the e2e code to use go:embed and yaml files
to create pod,pvc,and volumesnapshot resources. This makes
it easier to make changes when required.


e2e/pvc_check.go
```diff
@@ -11,9 +12,23 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-const (
-	size = "5Gi"
-)
+//go:embed testdata/pvc_tests/pvc-template.yaml
+var pvcYAMLTemplate string
```
sp98, line 16: `YAML` can be removed from each of the variables. Looks redundant IMO.  `pvcTemplate` should suffice.

e2e/utils.go
```diff
@@ -0,0 +1,33 @@
...
+	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
+	k8sv1 "k8s.io/api/core/v1"
+)
+
+func getPVC(pvcYAML string) (*k8sv1.PersistentVolumeClaim, error) {
+	obj, _, err := deserializer.Decode([]byte(pvcYAML), nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	pvc := obj.(*k8sv1.PersistentVolumeClaim)
```
sp98, line 13: I think `obj` can returned here without casting it to actual resource because the end resource is used in the `crClient.Create` function that can take the object.  
So we can have a single function that just decodes all the input yaml strings to obj.

sp98 (changes requested):

e2e/utils.go
nbalacha: I prefer casting it because it makes the code clearer as to what object it is working on.

e2e/testdata/pvc_tests/volume-snapshot.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshot
+metadata:
+  name: lvfilepvc-snapshot
+  namespace: lvm-endtoendtest
+spec:
+  volumeSnapshotClassName: odf-lvm-vg1
+  source:
+    persistentVolumeClaimName: lvmfilepvc
+
```
sp98, line 10: blank line.

e2e/pvc_check.go
nbalacha: I think the yaml makes it a little clearer as to what type it is in the code.

e2e/pvc_check.go
```diff
@@ -140,18 +173,27 @@ func pvcTest() {
...
 
+				By("Verifying that the Snapshot is ready")
+				Eventually(func() bool {
+					err := crClient.Get(ctx, types.NamespacedName{Name: snapshot.Name, Namespace: snapshot.Namespace}, snapshot)
+					return err == nil && *snapshot.Status.ReadyToUse
+				}, timeout, interval).Should(BeTrue())
+
 				By("Creating a clone of the block-pvc")
-				clonePvc = getSamplePvc(size, pvc.Name+"-clone", k8sv1.PersistentVolumeBlock, storageClass, "PersistentVolumeClaim", pvc.Name)
+				pvcCloneYaml := fmt.Sprintf(pvcCloneYAMLTemplate, "lvmblockpvc-clone", testNamespace, "Block", storageClass, "lvmblockpvc")
```
sp98, line 189: should we rather create multiple template files for reach resource type rather than updating the template like this?
I think the main purpose of the go embed would be to use static assets that don't change.

e2e/pvc_check.go
sp98: ok

e2e/utils.go
sp98: ok

e2e/pvc_check.go
nbalacha: I don't think that is necessary - we can use the same template to create multiple instances of a type. 
Having a separate file for each instance could quickly get out of hand once the tests are expanded.

sp98: ok

sp98 (approved):

PR #241, opened 2022-08-18
fix: check for multiple LVMCluster instances

nbalacha:
The operator requires that only a single instance of the
LVMCluster exists on the cluster.This fix fails the reconciliation
if the operator finds multiple LVMCluster instances on the cluster.


riya-singhal31:
The changes looks good.

riya-singhal31 (approved):

PR #243, opened 2022-08-18
fix: panic when checking if snapshot is ready

nbalacha:
This commit corrects the check for Snapshot.Status.ReadyToUse
and also reduces the interval at which the check is performed.


riya-singhal31:
The PR looks good.

riya-singhal31 (approved):

PR #278, opened 2022-10-12
fix: add test ns labels for PSA

nbalacha:
The PSA changes in OCP 4.12 prevent pods from starting as ns are labeled restricted by default. This commit sets the label pod-security.kubernetes.io/enforce: "baseline"
on the lvm-endtoendtest to allow the test pods to run. This also renames a file to conform to the naming convention.

### DIFF
--- a/e2e/helper.go
+++ b/e2e/helper.go
@@ -34,6 +34,9 @@ func createNamespace(ctx context.Context, namespace string) error {
 	label := make(map[string]string)
 	// label required for monitoring this namespace
 	label["openshift.io/cluster-monitoring"] = "true"
+	label["pod-security.kubernetes.io/enforce"] = "privileged"
+	label["security.openshift.io/scc.podSecurityLabelSync"] = "false"
+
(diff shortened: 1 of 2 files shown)


e2e/helper.go
```diff
@@ -34,6 +34,9 @@ func createNamespace(ctx context.Context, namespace string) error {
 	label := make(map[string]string)
 	// label required for monitoring this namespace
 	label["openshift.io/cluster-monitoring"] = "true"
+	label["pod-security.kubernetes.io/enforce"] = "baseline"
```
iamniting, line 37: Should not we use `privileged`

e2e/helper.go
```diff
@@ -52,6 +55,9 @@ func deleteNamespaceAndWait(ctx context.Context, namespace string) error {
...
```
iamniting, line 58: deleting the namespace should not require labels, Just a name should be enough.

iamniting (changes requested):

iamniting (approved):

PR #284, opened 2022-11-03
fix: add missing ns labels for make deploy

nbalacha:
The namespace created by make deploy was missing the PSA labels, preventing the topolvm-node and vg-manager pods from starting on OCP 4.12. This commit adds them and the missing cluster monitoring label.


iamniting (changes requested):
I think we should remove these `#`

nbalacha:
> I think we should remove these `#`

Fixed

iamniting (approved):

PR #290, opened 2022-11-09
chore: update the weekly build github action

nbalacha:
Updates the weekly build github action to call a second workflow. This will allow us to trigger a build for red-hat-storage/topolvm as part of the weekly build.

### DIFF
--- a/.github/workflows/manual-build.yaml
+++ b/.github/workflows/manual-build.yaml
@@ -1,34 +1,31 @@
-name: Manual Operator Build Dispatch
+name: manual-operator-build-push
 
 on:
+  workflow_call:
   workflow_dispatch:
     inputs:
       branch:
-        description: 'Branch to Build From'
+        description: 'Branch to build from'
         required: true
-      tag:
-        description: 'Additional tag for the build (such as alpha, beta, etc.) - Optional'
-        default: ''
-env:
-  GO_VERSION: "1.18"
-  IMG: "quay.io/lvmo/lvm-operator"
 
 jobs:
   build-and-publish-image:
     name: Build and publish the LVMO image
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v2
+      - uses: actions/checkout@v3
         with:
-          go-version: ${{ env.GO_VERSION }}
           ref: ${{ github.event.inputs.branch }}
-      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
 
       # 'test' is a sub-target of 'docker-build' and so cache go build & mod
       - id: go-cache-paths
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-build.yaml
```diff
@@ -1,34 +1,33 @@
...
     name: Build and publish the LVMO image
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: go.mod
           ref: ${{ github.event.inputs.branch }}
```
iamniting, line 23: I think this ref should go under the checkout not under setup-go

iamniting (changes requested):

.github/workflows/manual-build.yaml
```diff
@@ -45,21 +45,40 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: quay.io
-          username: ${{ secrets.QUAY_LVMO_ROBOT_NAME_2 }}
-          password: ${{ secrets.QUAY_LVMO_ROBOT_TOKEN_2 }}
+          username: ${{ secrets.QUAY_LVMO_ROBOT_NAME }}
```
iamniting, line 48: The changes look good to me, I think we should use `registry_namespace` here instead of the `secrets`. Because if someone gives values other than lvmo we won't be able to push the images though login will succeed. I would like it to fail earlier in that case at login itself. Also we can move login a bit up in the steps before go cache stuff.

iamniting (changes requested):

.github/workflows/manual-build.yaml
nbalacha: Fixed.

.github/workflows/weekly-build.yaml
```diff
@@ -1,51 +1,18 @@
...
-          password: ${{ secrets.QUAY_LVMO_ROBOT_TOKEN }}
-
-      - name: Build operator docker image
-        run: make docker-build-combined IMG=${{env.IMG}}
-
-      - name: Publish operator image
-        run: make docker-push IMG=${{env.IMG}}
+          workflow: manual-operator-build-push
+          repo: red-hat-storage/lvm-operator
+          inputs: '{ "branch": "main"}'
```
iamniting, line 18:
```suggestion
          inputs: '{ "branch": "main" }'
```

iamniting, line 18: or
```suggestion
          inputs: '{"branch": "main"}'
```

iamniting (changes requested):

iamniting (approved):
/lgtm
/approve

PR #299, opened 2022-12-08
fix: run weekly-build action only on the main repo instead of forks

iamniting:


iamniting:
Closing the PR after an offline discussion.

PR #300, opened 2022-12-12
Replace lvm-operator with lvms wherever required

nbalacha:
This changes the name of the lvm operator to lvms.


config/manifests/bases/lvm-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: operators.coreos.com/v1alpha1
```
iamniting, line 1: 2 CSV files are added, I think we should add only one.

iamniting (changes requested):

iamniting (approved):
/lgtm

PR #305, opened 2023-02-07
OCPVE-236: Reduce image size

suleymanakbas91:
This PR reduces the size of the images by
- specifying `--ldflags "-s -w"` in go build commands, which will strip debug information and remove DWARF symbols, reducing the size of the binary.
- changing the base image from `quay.io/centos/centos:stream8` to `registry.access.redhat.com/ubi8/ubi-minimal:8.7` on vgmanager and combined Dockerfiles.

The result is as follows:

Before: 
Screenshot 2023-01-26 at 17 04 40

After:
Screenshot 2023-02-14 at 19 52 16

The difference is big for vgmanager and combined images because of the base image change.

Signed-off-by: Suleyman Akbas <sakbas@redhat.com>

### DIFF
--- a/Dockerfile.combined
+++ b/Dockerfile.combined
@@ -17,15 +17,18 @@ COPY cmd/ cmd/
 COPY pkg/ pkg/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o vgmanager cmd/vgmanager/main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o metricsexporter cmd/metricsexporter/exporter.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build --ldflags "-s -w" -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build --ldflags "-s -w" -a -o vgmanager cmd/vgmanager/main.go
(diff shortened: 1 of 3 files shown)


jerpeter1:
Are we realizing most of the image size savings from using a different base image?

Dockerfile.combined
```diff
@@ -17,15 +17,18 @@ COPY cmd/ cmd/
...
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o vgmanager cmd/vgmanager/main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o metricsexporter cmd/metricsexporter/exporter.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build --ldflags "-s -w" -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build --ldflags "-s -w" -a -o vgmanager cmd/vgmanager/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build --ldflags "-s -w" -a -o metricsexporter cmd/metricsexporter/exporter.go
 
 # vgmanager needs 'nsenter' and other basic linux utils to correctly function
-FROM quay.io/centos/centos:stream8
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
jerpeter1, line 25: Do we want to specify an image digest, since this image is being built by an external team?

suleymanakbas91: Do you think it'd be ok specifying it as `latest` or should I use a specific version like `8.7.`?

jerpeter1: Specifying a specific version is generally preferred (although both are tags, so technically both are equally mutable). Best practice is to specify a specific image digest so that you have explicit control over when changes get incorporated into your own images and so it's easy to identify when those changes happened after the fact.

PR #307, opened 2023-02-14
OCPVE-253: Support multiple storage/device classes

suleymanakbas91:
This PR enables the support of multiple storage/device classes. 

- `api/v1alpha1/lvmcluster_types.go`: Introduces a new field `Default` in `DeviceClass`, which indicates whether the device-class is the default.
- `api/v1alpha1/lvmcluster_webhook.go`: Removes the validation for a single device class. Adds another validation to make sure there is only one default device class. If there is only one device class, it is considered the default one internally even if `Default` field is not set, to make the upgrades easier. Specifying more than one default device class errors TopoLVM.
- `api/v1alpha1/lvmvolumegroup_types.go`: Introduces a new field `Default` in `LVMVolumeGroupSpec` which matches the field in `DeviceClass` and is used internally. `controllers/lvm_volumegroup.go` change handles the matching.
- `pkg/vgmanager/vgmanager_controller.go`: Uses the new field `Default` in `LVMVolumeGroupSpec` to set `Default` for TopoLVM device class.

The rest of the code adjusts the tests and documentation for the new `Default` field.    

Signed-off-by: Suleyman Akbas <sakbas@redhat.com>

### DIFF
--- a/api/v1alpha1/lvmcluster_webhook.go
+++ b/api/v1alpha1/lvmcluster_webhook.go
@@ -48,11 +48,12 @@ func (l *LVMCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 func (l *LVMCluster) ValidateCreate() error {
 	lvmclusterlog.Info("validate create", "name", l.Name)
 
-	if len(l.Spec.Storage.DeviceClasses) != 1 {
-		return fmt.Errorf("Exactly one deviceClass is allowed")
+	err := l.verifySingleDefaultDeviceClass()
+	if err != nil {
+		return err
 	}
 
-	err := l.verifyPathsAreNotEmpty()
+	err = l.verifyPathsAreNotEmpty()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -164,6 +166,22 @@ func (l *LVMCluster) ValidateDelete() error {
...
+	if len(l.Spec.Storage.DeviceClasses) == 1 {
+		return nil
+	}
+	countDefault := 0
+	for _, deviceClass := range l.Spec.Storage.DeviceClasses {
+		if deviceClass.Default {
+			countDefault++
+		}
+	}
+	if countDefault != 1 {
```
jerpeter1, line 179: Is it possible for none of the DeviceClasses to be set Default (and therefore for countDefault to be 0)? If so, this would return a spurious error.

suleymanakbas91: I adjusted the function to return a different error message for each case.

PR #337, opened 2023-05-05
OCPVE-309: Remove commented code

qJkee:
Removed commented code 
Bumped ginkgo and controller-runtime


api/v1alpha1/webhook_suite_test.go
```diff
@@ -51,10 +50,6 @@ var cancel context.CancelFunc
 
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
-
-	RunSpecsWithDefaultAndCustomReporters(t,
```
suleymanakbas91, line 55: What does this do and why do we delete it? :)

test/e2e/helper.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/onsi/ginkgo/v2"
```
suleymanakbas91, line 24: I see that we use ginkgo v1 in the other files. Can we update them all? If it is a lot of work, then we can use v1 here as well to keep it consistent across the files.

test/e2e/helper.go
```diff
@@ -29,6 +30,11 @@ import (
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 )
 
+// debug immediately prints message in ginkgo test suite
+func debug(args ...interface{}) {
```
suleymanakbas91, line 34: Is this used somewhere?

PR #355, opened 2023-05-30
OCPVE-342: Change copyright in headers

qJkee:


suleymanakbas91:
Could you also update the reference in olm-deploy/lvms-catalogsource.yaml?

PR #356, opened 2023-05-31
OCPVE-349: Add PV and PVC controllers to publish events

suleymanakbas91:
This PR adds a PV and a PVC controller to the LVM Operator. 
- The PV controller watches for PVs created by LVMS and publishes an event if their `spec.claimRef` gets deleted: 

  Screenshot 2023-06-01 at 16 26 41

- The PVC controller watches for the PVCs created by LVMS and publishes an event if the PVC requests a storage greater than available capacity on any node:

  Screenshot 2023-06-06 at 12 22 11

### DIFF
--- a/controllers/persistent-volume-claim/controller.go
+++ b/controllers/persistent-volume-claim/controller.go
@@ -1,0 +1,135 @@
+package persistent_volume_claim
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"strings"
+
+	"github.com/openshift/lvm-operator/controllers"
+	"github.com/pkg/errors"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	capacityAnnotation = "capacity.topolvm.io/00default"
+)
+
+// PersistentVolumeClaimReconciler reconciles a PersistentVolumeClaim object
+type PersistentVolumeClaimReconciler struct {
+	client    client.Client
+	apiReader client.Reader
+	recorder  record.EventRecorder
+}
+
+// NewPersistentVolumeClaimReconciler returns PersistentVolumeClaimReconciler.
+func NewPersistentVolumeClaimReconciler(client client.Client, apiReader client.Reader, eventRecorder record.EventRecorder) *PersistentVolumeClaimReconciler {
+	return &PersistentVolumeClaimReconciler{
+		client:    client,
+		apiReader: apiReader,
+		recorder:  eventRecorder,
+	}
+}
+
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;update
+//+kubebuilder:rbac:groups=core,resources=node,verbs=get;list;watch
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;update;patch
+
+// Reconcile PVC
+func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.Log.WithName("persistentvolume-controller").WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
+
+	pvc := &corev1.PersistentVolumeClaim{}
+	err := r.client.Get(ctx, req.NamespacedName, pvc)
+	switch {
+	case err == nil:
+	case apierrors.IsNotFound(err):
+		return ctrl.Result{}, nil
+	default:
+		return ctrl.Result{}, err
+	}
+
+	// Skip if the PVC is deleted or does not use the lvms storage class.
+	if pvc.DeletionTimestamp != nil || !strings.HasPrefix(*pvc.Spec.StorageClassName, controllers.StorageClassPrefix) {
+		return ctrl.Result{}, nil
+	}
+
+	// Skip if the PVC is not in Pending state.
+	if pvc.Status.Phase != "Pending" {
+		return ctrl.Result{}, nil
+	}
+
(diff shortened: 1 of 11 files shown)


controllers/persistent-volume-claim/controller.go
```diff
@@ -0,0 +1,102 @@
...
+	found := false
+	for _, sc := range storageCapacityList.Items {
+		if strings.HasPrefix(sc.StorageClassName, controllers.StorageClassPrefix) && pvc.Spec.Resources.Requests.Storage().Cmp(*sc.Capacity) < 0 {
+			found = true
+		}
+	}
+
+	// Publish an event if the requested storage is greater than the available capacity
+	if !found {
+		r.recorder.Event(pvc, "Warning", "NotEnoughCapacity", "Requested storage is greater than available capacity on all the nodes. Check available capacity by `oc get csistoragecapacity`.")
```
brandisher, line 83:
```suggestion
		r.recorder.Event(pvc, "Warning", "NotEnoughCapacity", "Requested storage is greater than available capacity on any node. Check available node capacity by `oc get csistoragecapacity`.")
```
I'm not sure if this suggestion makes total sense but the idea is to note that the cluster itself may have storage available but the node where the storage is attempted to be provisioned does not.

Should this event include a reference to the node where the provisioning was tried? (assuming that information is available at this point in the reconcile) That might also help revise the message a little; something like "Requested storage for is greater than the available storage on `node-1`. Check available capacity by `oc get csistoragecapacity`"

suleymanakbas91: Kubernetes checks the capacity in all the nodes and then schedules the Pod referencing this PVC to a Node depending on the available storage. So, there is no node assignment yet. Instead, I'll update the message to have all the capacities of all the nodes: 
```
Requested storage (100000Gi) is greater than available capacity on any node (node ip-10-0-190-168.us-west-1.compute.internal has 2697273016320 bytes of free storage).
```
WDYT?

brandisher: I like it! From a useability point of view it makes it clearer what's being requested and provides evidence that none of the nodes can service the request. One nit, can the node's available space unit match the requested space unit? (e.g. if requested is `Gi` then node space available is also `Gi`)

This will be helpful in the future when we start allowing less than gigabyte sized requests and makes the quick visual comparison easier.

controllers/persistent-volume/controller.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	// Skip if the PV is deleted or PV does not use the lvms storage class.
+	if pv.DeletionTimestamp != nil || !strings.HasPrefix(pv.Spec.StorageClassName, controllers.StorageClassPrefix) {
+		return ctrl.Result{}, nil
+	}
+
+	// Publish an event if PV has no claimRef
+	if pv.Spec.ClaimRef == nil {
+		r.recorder.Event(pv, "Warning", "RemovedClaimReference", "Claim reference is removed. This PV is dynamically managed. Please do not modify it.")
```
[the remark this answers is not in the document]
suleymanakbas91: It would fix this, but there is no way to know if the claimRef was already there or re-added.

controllers/persistent-volume-claim/controller.go
suleymanakbas91: I've added a simple function to pretty-print the size: https://github.com/openshift/lvm-operator/compare/f2975147ae757cd8cb27be323b65ae2b14231582..053df2bb1d1d852644faa73d1547b4b94ca50d68

controllers/persistent-volume/controller.go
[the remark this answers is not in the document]
suleymanakbas91: I don't think it'd be possible to add a claimRef with a reference to a non-existing PVC, but I'll test this.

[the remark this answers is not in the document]
suleymanakbas91: I'll also test if it's possible to add a claimRef with a reference to a PVC with another storage class than LVMS one.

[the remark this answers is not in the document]
suleymanakbas91: For both scenarios, TopoLVM gives me the same error: 
```
Logical volume vg1/3b359222-d9c5-45ce-b806-9cada1646085 contains a filesystem in use.
2023-06-05T16:32:56.735261Z topolvm-node-n5779 lvmd error: "failed to remove volume" error="exit status 5" name="3b359222-d9c5-45ce-b806-9cada1646085"
```
It only works if I add back the exact same claimRef. Other than that, it does not delete the underlying volume so the "critical data" is safe.

main.go
```diff
@@ -37,6 +37,8 @@ import (
 
 	lvmv1alpha1 "github.com/openshift/lvm-operator/api/v1alpha1"
 	"github.com/openshift/lvm-operator/controllers"
+	persistent_volume "github.com/openshift/lvm-operator/controllers/persistent-volume"
+	persistent_volume_claim "github.com/openshift/lvm-operator/controllers/persistent-volume-claim"
```
brandisher, line 41: Same here. Any reason not to use `pvc` instead of `persistent_volume_claim`?

pkg/internal/block_device.go
```diff
@@ -89,42 +86,17 @@ func ListBlockDevices(exec Executor) ([]BlockDevice, error) {
 	return blockDeviceMap["blockdevices"], nil
 }
 
-func ListDiskByPath(exec Executor) (map[string]string, error) {
```
brandisher, line 92: Is this a cleanup of something that's not used?

brandisher:
Couple of questions/nits but lgtm. Keeping the hold on so @jeff-roche can take a look as well.

/lgtm

main.go
```diff
@@ -37,6 +37,8 @@ import (
 
 	lvmv1alpha1 "github.com/openshift/lvm-operator/api/v1alpha1"
 	"github.com/openshift/lvm-operator/controllers"
+	persistent_volume "github.com/openshift/lvm-operator/controllers/persistent-volume"
```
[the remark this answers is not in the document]
suleymanakbas91: `pv` also means physical volumes in this context, so I try to keep it clear by writing the full names.

main.go
suleymanakbas91: For the same reason above, I will keep it as `persistent_volume_claim`.

pkg/internal/block_device.go
suleymanakbas91: Yes, this was a leftover code that is not used anywhere so I deleted it.

suleymanakbas91:
@jeff-roche was only waiting for the claimRef test before giving his approval, so I think it should be ok to unhold it now.

/unhold

PR #357, opened 2023-06-02
OCPVE-302: feat: fstype e2e tests

jeff-roche:


test/e2e/lvmcluster.go
```diff
@@ -84,3 +93,106 @@ func deleteLVMCluster(ctx context.Context) error {
...
+				}
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			Expect(sc.Parameters["csi.storage.k8s.io/fstype"]).To(Equal(string(v1alpha1.FilesystemTypeXFS)))
+		})
+
+		It("should be xfs if specified", func() {
+			clusterConfig = generateLVMCluster()
+			clusterConfig.Spec.Storage.DeviceClasses[0].FilesystemType = v1alpha1.FilesystemTypeXFS
```
brandisher, line 158: Question for my own edification...why is `DeviceClasses[0]` selected here but not in the default case? Is it just the case where if the user sets it specifically they must have set `spec.Storage.DeviceClasses[0]`?

brandisher:
One minor question but looks good otherwise. Holding for @qJkee's review.

/lgtm
/hold

test/e2e/lvmcluster.go
jeff-roche: Yeah they could in theory set multiple device classes. Each device class will have a default value of XFS if one isn't specified. So this is a test case where it would be specified as XFS rather than using the default value

PR #360, opened 2023-06-23
OCPVE-330: Add add owner reference to created objects

qJkee:
Add owner reference to created objects. CR owns all created objects for given cluster


jeff-roche:
Confirmed tests by running manually
/override ci/prow/lvm-operator-bundle-e2e-aws

PR #366, opened 2023-07-20
OCPVE-406: fix: updated a few issues with the cluster build commands

jeff-roche:


jeff-roche:
override ci/prow/lvm-operator-bundle-e2e-aws

PR #368, opened 2023-07-21
USHIFT-1059: fix: setup microshift kustomization target for use in microshift rebases

jakobmoellerdev:
avoids missing synced RBACs that have now gone into SCV. In general, this needs to be removed at some point and unified with the actual RBACs used by LVM operator.

Needs to be verified with rebase dry run before it can be merged


jakobmoellerdev:
Closing as change was made to extract CSV instead so this PR is no longer necessary: https://github.com/openshift/microshift/pull/2107

PR #369, opened 2023-07-26
OCPBUGS-15576: fix: ensure panic safety in PVC controller for non set storageClassName

jakobmoellerdev:
Otherwise if a PVC is created without storageClassName, LVM Operator will crash and cause a panic.

Also removes unnecessary API Reader in PVC controller and cleans up logging, introduces test cases for ignore cases

### DIFF
--- a/controllers/persistent-volume-claim/controller.go
+++ b/controllers/persistent-volume-claim/controller.go
@@ -26,17 +26,15 @@ const (
 
 // PersistentVolumeClaimReconciler reconciles a PersistentVolumeClaim object
 type PersistentVolumeClaimReconciler struct {
-	client    client.Client
-	apiReader client.Reader
-	recorder  record.EventRecorder
+	Client   client.Client
+	Recorder record.EventRecorder
 }
 
 // NewPersistentVolumeClaimReconciler returns PersistentVolumeClaimReconciler.
-func NewPersistentVolumeClaimReconciler(client client.Client, apiReader client.Reader, eventRecorder record.EventRecorder) *PersistentVolumeClaimReconciler {
+func NewPersistentVolumeClaimReconciler(client client.Client, eventRecorder record.EventRecorder) *PersistentVolumeClaimReconciler {
 	return &PersistentVolumeClaimReconciler{
-		client:    client,
-		apiReader: apiReader,
-		recorder:  eventRecorder,
+		Client:   client,
+		Recorder: eventRecorder,
 	}
 }
 
@@ -46,10 +44,10 @@ func NewPersistentVolumeClaimReconciler(client client.Client, apiReader client.R
 
 // Reconcile PVC
 func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.Log.WithName("persistentvolume-controller").WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
+	logger := log.Log.WithName("pvc-controller").WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
 
 	pvc := &corev1.PersistentVolumeClaim{}
-	err := r.client.Get(ctx, req.NamespacedName, pvc)
+	err := r.Client.Get(ctx, req.NamespacedName, pvc)
 	switch {
 	case err == nil:
 	case apierrors.IsNotFound(err):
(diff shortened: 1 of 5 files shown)


controllers/persistent-volume-claim/controller.go
```diff
@@ -26,17 +26,15 @@ const (
 
 // PersistentVolumeClaimReconciler reconciles a PersistentVolumeClaim object
 type PersistentVolumeClaimReconciler struct {
-	client    client.Client
-	apiReader client.Reader
-	recorder  record.EventRecorder
+	Client   client.Client
```
jeff-roche, line 29: Why the scope change? And more importantly, why are we removing the apiReader?

controllers/persistent-volume-claim/controller.go
```diff
@@ -46,10 +44,10 @@ func NewPersistentVolumeClaimReconciler(client client.Client, apiReader client.R
 
 // Reconcile PVC
 func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := log.Log.WithName("persistentvolume-controller").WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
+	logger := log.Log.WithName("persistentvolumeclaim-controller").WithValues("Request.Name", req.Name, "Request.Namespace", req.Namespace)
```
jeff-roche, line 47: I like this name change, we have too much name overlap as it is so clarifying where we can is a good thing

controllers/persistent-volume-claim/controller_test.go
jeff-roche, line 1: 🥲

controllers/persistent-volume-claim/controller_test.go
```diff
@@ -0,0 +1,163 @@
...
+				Spec:       v1.PersistentVolumeClaimSpec{},
+				Status:     v1.PersistentVolumeClaimStatus{},
+			}),
+			want: controllerruntime.Result{
+				Requeue:      false,
+				RequeueAfter: 0,
+			},
+		},
+		{
+			name: "testing non-applicable storageClassName",
```
jeff-roche, line 66: I wouldn't do non-applicable here, I'd just say `"testing empty storageClassName"`

controllers/persistent-volume-claim/controller.go
jakobmoellerdev: apiReader was unused.
Scope Change as tests are in "persistent_volume_claim_test" package as per Go recommendations, and to inject fields without Getter/Setter I had to export them to be accesible in the test package

controllers/persistent-volume-claim/controller_test.go
jakobmoellerdev: its not empty though, the empty case is already covered (see above). This case is for a storage class name that is not empty, but also not prefixed with `controllers.StorageClassPrefix`.

jakobmoellerdev: Just to be safe I also changed the object names in the test cases to make the distinction clearer, hope that resolves the uncertainty

controllers/persistent-volume-claim/controller_test.go
```diff
@@ -0,0 +1,163 @@
...
+				Name:      "test-emptyStorageClassName",
+			}},
+			client: fake.NewClientBuilder().WithObjects(&v1.PersistentVolumeClaim{
+				TypeMeta:   metav1.TypeMeta{},
+				ObjectMeta: metav1.ObjectMeta{Namespace: defaultNamespace, Name: "test-emptyStorageClassName"},
+				Spec:       v1.PersistentVolumeClaimSpec{},
+				Status:     v1.PersistentVolumeClaimStatus{},
+			}),
+			want: controllerruntime.Result{
+				Requeue:      false,
```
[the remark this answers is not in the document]
jakobmoellerdev: originally thought for positive test cases later on with different results but removed for now to make it smaller

jakobmoellerdev:
@jeff-roche e2e test is still flaky right? Should we wait for them to be fixed or do you want to force this merge in?

jeff-roche:
> @jeff-roche e2e test is still flaky right? Should we wait for them to be fixed or do you want to force this merge in?

I will run the e2e tests manually and report back

jeff-roche:
Manually tested
/override ci/prow/lvm-operator-bundle-e2e-aws

jeff-roche:
Manually tested
/override ci/prow/lvm-operator-bundle-e2e-aws

PR #378, opened 2023-08-08
OCPVE-619: Refactor E2E Tests to be context aware

jakobmoellerdev:
Also introduces flag to skip Snapshot Tests on clusters without Snapshot CRDs by default (Openshift Local). 

With the given guide and tests, I can get the entire snapshot test suite to run locally in about 10 minutes Right now. The rest of the latency fixes has to come from a follow-up PR that investigates minDeviceAge (opened up [OCPVE-622](https://issues.redhat.com/browse/OCPVE-622) for this). After some testing, I believe we could get the test runtime down to sub 5 minutes with some small optimizations.

Known Issues:
In PVC and ephemeral tests the cleanup is run as a test. This causes leftover resources when the test fails and the suite cannot cleanup properly. This would require a bigger rewrite to the tests to fix so Im leaving this out.

Also:
- Bumps Ginkgo CLI tool download to 2.9.5 to avoid warning of mismatch between go and cli version
- Fixes Kubeconfig Lookup in E2E Tests as it was not returning the error in case the Kubeconfig was not found, now its displayed with `fmt.Errorf`
- Fixes Klog Crashing in Tests because logger was not redirected correctly
- Fixes Seccomp Profile on test pods which causes warnings during creation during tests in KubeAPI Warning Logger
- Moves Testing Timeout from 15 minutes to 2 minutes
- Moves Testing Polling Interval from 15 seconds to 3 seconds
- Changes Deletion Pending Check interval from 1 Minute to 10 seconds
- Introduces a E2E testing documentation with OpenShift Local that is easy to follow and understand, also explains how to enable CRC containers for snapshot and cloning support (not supported out of the box, one has to deploy volume snapshot capabilities themselves)
- Moves `HACKING.md` into the docs folder and combines it with the E2E guide


test/e2e/validation.go
```diff
@@ -43,159 +46,112 @@ const (
...
-			debug("Error getting TopoLVM node daemonset %s: %s\n", topolvmNodeDaemonSetName, err.Error())
-		}
-		return ds.Status.DesiredNumberScheduled == ds.Status.NumberReady
-	}, timeout, interval).Should(BeTrue())
-	debug("TopoLVM node pods are ready\n")
-
-	return nil
+func validateTopolvmNode(ctx context.Context) bool {
+	By("validating the TopoLVM Node DaemonSet")
+	return validateDaemonSet(ctx, types.NamespacedName{Name: topolvmNodeDaemonSetName, Namespace: installNamespace})
```
jeff-roche, line 88: What's the point of breaking the business logic here out into another function?

test/e2e/validation.go
```diff
@@ -43,159 +46,112 @@ const (
...
-			debug("Error getting vg manager daemonset %s: %s\n", vgManagerDaemonsetName, err.Error())
-		}
-		return err == nil
-	}, timeout, interval).Should(BeTrue())
-	debug("VG manager daemonset found\n")
-
-	return nil
+func validateVGManager(ctx context.Context) bool {
+	By("validating the vg-manager DaemonSet")
+	return validateDaemonSet(ctx, types.NamespacedName{Name: vgManagerDaemonsetName, Namespace: installNamespace})
```
jeff-roche, line 94: Same comment about breaking it out

test/e2e/validation.go
```diff
@@ -43,159 +46,112 @@ const (
...
 }
 
 // function to validate TopoLVM Deployment.
-func validateTopolvmController(ctx context.Context) error {
-	dep := appsv1.Deployment{}
+func validateTopolvmController(ctx context.Context) bool {
+	By("validating the TopoLVM controller deployment")
+	return validateDeployment(ctx, types.NamespacedName{Name: topolvmCtrlDeploymentName, Namespace: installNamespace})
```
jeff-roche, line 100: And again about breaking it out

test/e2e/validation.go
```diff
@@ -43,159 +46,112 @@ const (
...
+		if !isReady {
+			return fmt.Errorf("the DaemonSet %s is not considered ready", name)
+		}
+		return nil
+	}, timeout, interval).WithContext(ctx).Should(Succeed())
+}
 
-	debug("TopoLVM controller deployment found\n")
-	return nil
+func validateDeployment(ctx context.Context, name types.NamespacedName) bool {
```
jeff-roche, line 117: I know it's returning the result of the eventually but I'd still rather see this in the above functions and reduce the amount of hopping around we need to do in the code. In typical code I'd be more okay with it if it made it easier to unit test but that isn't the case here since we are in e2e

test/e2e/validation.go
jakobmoellerdev: Code duplication is never good, and both topoLVM and VG Manager share the same validation code for the daemon set readiness.

test/e2e/validation.go
jakobmoellerdev: This one is probably not really necessary, will remove again

test/e2e/validation.go
jakobmoellerdev: in general I agree hopping is bad, but currently there is a clear layer now where something fails because I passed through the ginkgo context. It will give you an exact stack trace showing you where things failed with each call pass being shown as well. If you still think we should move it out I will do so of course!

docs/debugging/local-e2e-guide.md
jeff-roche, line 1: Can you either combine this one with HACKING.md or alternatively bring HACKING.md into this folder? I this file and that file make sense to live together. Another thing you could do is link to this one from HACKING.md.

suleymanakbas91 (changes requested):

suleymanakbas91 (changes requested):

test/e2e/lvmcluster.go
```diff
@@ -126,82 +127,91 @@ func lvmClusterTest() {
 	Describe("Filesystem Type", Serial, func() {
 
 		var clusterConfig *v1alpha1.LVMCluster
-		ctx := context.Background()
```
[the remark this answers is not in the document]
jakobmoellerdev: Good catch. Actually these methods were never used (thats why I didnt catch it) so i simply removed them.

controllers/topolvm_snapshotclass.go
```diff
@@ -72,10 +75,16 @@ func (s topolvmVolumeSnapshotClass) ensureDeleted(r *LVMClusterReconciler, ctx c
 
 		if err != nil {
 			// already deleted in previous reconcile
-			if errors.IsNotFound(err) {
+			if k8serrors.IsNotFound(err) {
 				r.Log.Info("topolvm volume snapshot class is deleted", "VolumeSnapshotClass", vscName)
 				return nil
 			}
+			if runtime.IsNotRegisteredError(err) || meta.IsNoMatchError(err) ||
```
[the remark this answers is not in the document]
jakobmoellerdev: Yes you are completely right. This didnt get catched because a failure will actually not change the LVMCluster to failed it will just log "failed to reconcile". This has to be reflected properly in the status, but that will be tackled with https://issues.redhat.com/browse/OCPVE-627 when we revisit the resource creation / status checks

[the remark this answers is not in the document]
jakobmoellerdev: I anyhow added the check now also to create obviously

PR #380, opened 2023-08-11
OCPVE-622: chore: remove deviceMinAge due to imperfect validation

jakobmoellerdev:
This removes deviceMinAge as it is not a reliable way to filter devices created by other entities.

Case to cover:
Device is freshly attached by another entity and should not be picked up by LVM Operator

Reason it doesn't work:
1. Device is attached at t0
2. Device is looked up LVM Operator at t1
3. External entity does not lookup Device in t0+30s
4. Device is viewed as available by LVM operator at t1+30s and gets used anyhow

On top of this, if a device gets created and used by LVM, LVM puts a system lock in place so there cannot ever be concurrent calls to LVM. This makes this condition uninteresting and we can remove minDeviceAge


pkg/vgmanager/vgmanager_controller.go
```diff
@@ -263,11 +244,7 @@ func (r *VGReconciler) reconcile(ctx context.Context, volumeGroup *lvmv1alpha1.L
 	}
 
 	// requeue faster if some devices are too recently observed to consume
```
suleymanakbas91, line 246: This line can also be deleted.

suleymanakbas91 (changes requested):

PR #382, opened 2023-08-14
OCPVE-625: docs: Add RAID limitation and workaround (for LVM Software Raid)

jakobmoellerdev:
While we did some investigations into the topic around RAID it became apparent that LVM Software RAID cannot be supported reliably which is why we should add a new limitation here. In general we should also add another Consideration on never using software RAID in the first place if possible, but thats another issue so I am leaving this out of here.


README.md
```diff
@@ -380,6 +380,24 @@ LVMS does not support the reconciliation of multiple LVMCluster custom resources
 
 It is not possible to upgrade from release-4.10 and release-4.11 to a newer version due to a breaking change that has been implemented. For further information on this matter, consult [the relevant documentation](https://github.com/topolvm/topolvm/blob/main/docs/proposals/rename-group.md).
 
+### Missing native LVM RAID Configuration support
```
suleymanakbas91, line 383: Could you please update the table of contents as well?

suleymanakbas91 (changes requested):

suleymanakbas91:
It looks pretty good from the engineering side, but someone from the docs team should also check for the language, formatting, etc. That's why I'm adding a hold. 
/lgtm
/approve
/hold

PR #384, opened 2023-08-16
OCPVE-620: docs: add guidance on Multi-Node Topology Snapshotting

jakobmoellerdev:
Creates a guidance documentation on Snapshotting and Cloning in Multi-Node Topologies. It specifically recommends to use Stateful Sets alongside preprovisioned and bound source PVCs to avoid running into scheduler limitations that may (or may not) occur in TopoLVM CSI scheduling (I opened a discussion in https://github.com/topolvm/topolvm/discussions/737 to clarify the exact conditions, also I am trying to independently verify this issue).


suleymanakbas91:
Great explanation!
/lgtm
/approve
/hold 
for docs team to review.

PR #385, opened 2023-08-16
OCPBUGS-17805: Do not error during capacity check if annotation does not exist on the node

suleymanakbas91:
If the capacity annotation (`capacity.topolvm.io/deviceClass`) does not exist on a node, that means the node is not used by LVMS. So, do not error, but just do not include it in the calculations during the capacity check.  

Change the implementation to use the corresponding annotation for the device class instead of using the default annotation.

### DIFF
--- a/controllers/persistent-volume-claim/controller.go
+++ b/controllers/persistent-volume-claim/controller.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	capacityAnnotation = "capacity.topolvm.io/00default"
+	CapacityAnnotation = "capacity.topolvm.io/"
 )
 
 // PersistentVolumeClaimReconciler reconciles a PersistentVolumeClaim object
@@ -67,7 +67,9 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
-	if !strings.HasPrefix(*pvc.Spec.StorageClassName, controllers.StorageClassPrefix) {
+	// Skip if StorageClassName does not contain the lvms prefix
+	lvmsPrefix, deviceClass, exists := strings.Cut(*pvc.Spec.StorageClassName, "-")
+	if !exists || fmt.Sprintf("%s-", lvmsPrefix) != controllers.StorageClassPrefix {
 		logger.Info("skipping pvc as the storageClassName does not contain desired prefix",
 			"desired-prefix", controllers.StorageClassPrefix)
 		return ctrl.Result{}, nil
@@ -90,11 +92,9 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 	requestedStorage := pvc.Spec.Resources.Requests.Storage()
 	var nodeMessage []string
 	for _, node := range nodeList.Items {
-		capacity, ok := node.Annotations[capacityAnnotation]
+		capacity, ok := node.Annotations[CapacityAnnotation+deviceClass]
 		if !ok {
-			errMessage := fmt.Sprintf("could not find capacity annotation on the node %s", node.Name)
-			logger.Error(fmt.Errorf(errMessage), errMessage)
-			return ctrl.Result{}, nil
+			continue
 		}
 
 		capacityQuantity, err := resource.ParseQuantity(capacity)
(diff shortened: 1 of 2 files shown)


controllers/persistent-volume-claim/controller.go
```diff
@@ -92,9 +92,7 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 	for _, node := range nodeList.Items {
 		capacity, ok := node.Annotations[capacityAnnotation]
 		if !ok {
-			errMessage := fmt.Sprintf("could not find capacity annotation on the node %s", node.Name)
-			logger.Error(fmt.Errorf(errMessage), errMessage)
-			return ctrl.Result{}, nil
+			continue
```
jakobmoellerdev, line 95: Could we write a test case for the controller not erroring?

jakobmoellerdev: On that same note, couldn't it happen that there is more than one deviceClass present on TopoLVM side? Then this annotation would increment +1, right?

suleymanakbas91: I was just thinking about the multiple deviceClasses case where we would actually need to check the capacity label corresponding to the storage class that is specified in the requesting pvc.

suleymanakbas91: There are multiple annotations based on the number of deviceClasses. The annotation I'm using here is showing the capacity of the default deviceClass.

jakobmoellerdev: Im trying to understand if its possible to have a non-default deviceClass be relevant here. Afaik we controll the lvmd config from lvm operator so we should never come into a non-default case right?

suleymanakbas91: Topolvm errors if there is no single default deviceClass specified in lvmd config. However, here it is not much relevant. 
- Let's say the user specified two deviceClasses (`vg1` and `vg2`). 
- In this case, lvms creates two storageClasses (`lvms-vg1` and `lvms-vg2`)
- Also, Topolvm adds a capacity label to each node for both deviceClasses (`capacity.topolvm.io/vg1: 96427048960`, `capacity.topolvm.io/vg2: 192979927040`), and one for the default deviceClass which is vg1 in this case (`capacity.topolvm.io/00default: 96427048960`). 

In this implementation of capacity warning, we should check which storageClass is being used in the pvc and print the capacity for that storageClass/deviceClass using the capacity annotations. This is irrelevant to whether deviceClass is the default or not.

jakobmoellerdev: Then in this case the PVC check here needs to change to accommodate for the correct capacity annotation I believe? Since if we check for default but another deviceClass would be used it would be irrelevant for the PVC?

suleymanakbas91: Exactly. I'll update the implementation.

controllers/persistent-volume-claim/controller.go
```diff
@@ -67,7 +67,9 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
-	if !strings.HasPrefix(*pvc.Spec.StorageClassName, controllers.StorageClassPrefix) {
+	// Skip if StorageClassName does not contain the lvms prefix
+	lvmsPrefix, deviceClass, exists := strings.Cut(*pvc.Spec.StorageClassName, "-")
```
jakobmoellerdev, line 71: If StorageClassName is not set this will panic. Empty StorageClassNames are allowed so this could be problematic without a check.

jakobmoellerdev: Maybe we should add some smarter way in a separate follow-up issue (e.g. a label or annotation) instead of this prefix thing as an issue to differentiate the correct storageclass in the future. WDYT?

jakobmoellerdev (changes requested):

controllers/persistent-volume-claim/controller.go
suleymanakbas91: I agree. I'm also not a fan of using string operations to figure this out :)

suleymanakbas91: There is a parameter in the storageclass specifying the deviceclass: 
`Parameters:            csi.storage.k8s.io/fstype=xfs,topolvm.io/device-class=vg1`.

So, I could get the storageclass object and check the parameter value to obtain the deviceclass. Wdyt? Would it be an overkill?

jakobmoellerdev (approved):
/lgtm

Good stuff :)

PR #387, opened 2023-08-17
OCPBUGS-17848: fix: don't fail if devices already in vg

jeff-roche:
It was discovered as part of the 4.14 verification process that the available devices check was creating an error due to not properly skipping devices that were already part of the volume group. This error would lead to a degraded state of the lvmcluster when the first reconciliation loop was initiated.


jakobmoellerdev (approved):
/lgtm

PR #392, opened 2023-08-22
OCPVE-361: chore: allow multi-arch with ARCH / OS Flag

jakobmoellerdev:
Allows building multi-arch images with ARCH / OS Flag from makefile. Also adjusts dockerfile to have the correct flags passed to the builder and running platform image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,20 +169,22 @@ run: manifests generate ## Run the Operator from your host.
 ##@ Build
 
 IMAGE_BUILD_CMD ?= $(shell command -v docker 2>&1 >/dev/null && echo docker || echo podman)
+OS ?= linux
+ARCH ?= amd64
 
 all: build
 
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager main.go
+	GOOS=$(OS) GOARCH=$(ARCH) go build -o bin/manager main.go
 
 build-vgmanager: generate fmt vet ## Build vg manager binary.
-	go build -o bin/vgmanager cmd/vgmanager/main.go
+	GOOS=$(OS) GOARCH=$(ARCH) go build -o bin/vgmanager cmd/vgmanager/main.go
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -169,20 +169,22 @@ run: manifests generate ## Run the Operator from your host.
...
 build-vgmanager: generate fmt vet ## Build vg manager binary.
-	go build -o bin/vgmanager cmd/vgmanager/main.go
+	GOOS=$(OS) GOARCH=$(ARCH) go build -o bin/vgmanager cmd/vgmanager/main.go
 
 build-prometheus-alert-rules: jsonnet monitoring/mixin.libsonnet monitoring/alerts/alerts.jsonnet monitoring/alerts/*.libsonnet
 	$(JSONNET) -S monitoring/alerts/alerts.jsonnet > config/prometheus/prometheus_rules.yaml
 
 docker-build: ## Build docker image with the manager.
-	$(IMAGE_BUILD_CMD) build -t ${IMG} .
+	$(IMAGE_BUILD_CMD) build --platform=${OS}/${ARCH} -t ${IMG} .
```
suleymanakbas91, line 187: This does not work on darwin/arm64: 

```
➜  ✗ make docker-build
podman build --platform=darwin/arm64 -t quay.io/test/lvms-operator:tag .
[1/2] STEP 1/13: FROM golang:1.20 AS builder
Resolving "golang" using unqualified-search registries (/etc/containers/registries.conf.d/999-podman-machine.conf)
Trying to pull docker.io/library/golang:1.20...
Error: creating build container: choosing an image from manifest list docker://golang:1.20: no image found in manifest list for architecture arm64, variant "", OS darwin

make: *** [docker-build] Error 125
```
I don't know if we will ever need to build container images for different platforms from here. Can we make this at least parametric, meaning it will build for linux/amd64 by default unless related parameters are specified?

suleymanakbas91 (changes requested):

Makefile
jakobmoellerdev: So it was actually parametric already, but used `go env` variables for OS / ARCH. I just hardcoded the Args to `linux/amd64` now. You should be able to run this normally without any changes now.

PR #395, opened 2023-08-23
feat: switch to SSA for resource reconciliation

jakobmoellerdev:
This is a sample implementation for SSA within lvm-operator and changes resource synchronization to no longer use Create/Update but use the k8s API Server for conflict resolution


jakobmoellerdev:
Closing due to too many open variables. Once a few more issues are closed and Applyconfiguration is available for kubebuilder based controllers I think we can give it another shot.

PR #396, opened 2023-08-24
OCPVE-667: fix: allow no default deviceClass to be set

jakobmoellerdev:
This allows no default deviceClass to be present. Since this is a change from previous validation it now includes a warning if no default is set: "no default deviceClass was specified, it will be mandatory to specify the storage class"

TODO:

- [x] Test case using no default deviceClass that succeeds in e2e tests

### DIFF
--- a/api/v1alpha1/lvmcluster_webhook.go
+++ b/api/v1alpha1/lvmcluster_webhook.go
@@ -48,67 +48,71 @@ func (l *LVMCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (l *LVMCluster) ValidateCreate() (admission.Warnings, error) {
 	lvmclusterlog.Info("validate create", "name", l.Name)
+	warnings := admission.Warnings{}
 
-	err := l.verifySingleDefaultDeviceClass()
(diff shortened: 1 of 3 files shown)


api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -197,26 +201,25 @@ func (l *LVMCluster) ValidateDelete() (admission.Warnings, error) {
...
 	}
-	if countDefault < 1 {
-		return fmt.Errorf("one default deviceClass is required. Please specify default=true for the default deviceClass")
-	} else if countDefault > 1 {
-		return fmt.Errorf("only one default deviceClass is allowed. Currently, there are %d default deviceClasses", countDefault)
+	if countDefault > 1 {
+		return nil, fmt.Errorf("only one default deviceClass is allowed. Currently, there are %d default deviceClasses", countDefault)
+	}
+	if countDefault == 0 {
+		return admission.Warnings{"no default deviceClass was specified, it will be mandatory to specify the storage class"}, nil
```
suleymanakbas91, line 219: I think we should clarify where it will be mandatory to specify the storage class. Something like `.. the storage class in the PVCs`.

suleymanakbas91 (changes requested):

PR #399, opened 2023-08-28
OCPVE-664: chore: refactor SCC creation and deletion in LVM Operator

jakobmoellerdev:
This commit refactors the SecurityContextConstraint (SCC) creation and deletion in the LVM Operator. Previously, the check to determine if the SCC should be created or deleted was performed on every reconcile loop. This resulted in unnecessary API calls on non-Openshift clusters.

With this commit, the check is performed only once, during operator start-up and the information is stored in the LVMClusterReconciler structure. Therefore, the SCC resource is only manipulated if the operator is running on an Openshift cluster, saving resources on other cluster types.

This refactor makes the operator more efficient and potentially reduces the response time of reconcile loop when running on non-Openshift clusters.


controllers/lvmcluster_controller.go
```diff
@@ -65,19 +63,12 @@ type resourceManager interface {
...
-)
-
 // LVMClusterReconciler reconciles a LVMCluster object
 type LVMClusterReconciler struct {
 	client.Client
 	Scheme         *runtime.Scheme
 	Log            logr.Logger
-	ClusterType    ClusterType
+	ClusterType    cluster.Type
 	SecurityClient secv1client.SecurityV1Interface
```
suleymanakbas91, line 72: Is this `SecurityClient` now used anywhere?

jakobmoellerdev: No I completely removed it from the sources, and replaced it with the canonical client-go plus scheme registration

jakobmoellerdev:
/hold
still investigating how we best go about the informer creation and watches for SCCs

PR #401, opened 2023-08-28
OCPVE-399: test: validate snapshot content

jakobmoellerdev:
Adds test logic to validate content that is written to the volume managed by TopoLVM and cloned / snapshotted


test/e2e/testdata/pvc_tests/pod-volume-mount-template.yaml
```diff
@@ -5,15 +5,16 @@ metadata:
...
+    - name: ubuntu
       securityContext:
         runAsNonRoot: true
         seccompProfile:
           type: RuntimeDefault
         capabilities:
           drop:
             - "ALL"
-      image: public.ecr.aws/docker/library/nginx:latest
+      image: quay.io/cybozu/ubuntu:20.04
```
suleymanakbas91, line 16: I see that TopoLVM switched to `registry.k8s.io/pause` instead: https://github.com/topolvm/topolvm/pull/621/files

jakobmoellerdev: Interesting find, adopted the images, but will have to see if the e2e tests run on them with exec

jakobmoellerdev: changed them to buysbox on ECR instead as we need a shell for interacting with the data. Still lots smaller than the cybozu ubuntu images though

PR #402, opened 2023-08-30
OCPVE-670: fix ownerref alignment and deletion logic on removal

jakobmoellerdev:
Deletion Logic was revamped for all the resourceManagers. also fixes up various log messages to have a consistent form during deletion. 
Also adjusts the unit tests to correctly work with ownerref removal. Also makes sure that the Owner references are always set in the create or update logic since otherwise they would not be updated.

Removes ensureDeleted for `controllers/topolvm_node.go` as well as `controllers/topolvm_controller.go`

Also kills multierror and its only call in favor or errors.Join


controllers/vgmanager.go
```diff
@@ -62,6 +56,9 @@ func (v vgManager) ensureCreated(r *LVMClusterReconciler, ctx context.Context, l
 	// the anonymous mutate function modifies the daemonset object after fetching it.
 	// if the daemonset does not already exist, it creates it, otherwise, it updates it
 	result, err := ctrl.CreateOrUpdate(ctx, r.Client, ds, func() error {
+		if err := ctrl.SetControllerReference(lvmCluster, ds, r.Scheme); err != nil {
```
suleymanakbas91, line 59: It is already set in the line 87.

controllers/topolvm_controller.go
```diff
@@ -84,37 +81,12 @@ func (c topolvmController) ensureCreated(r *LVMClusterReconciler, ctx context.Co
...
-		return fmt.Errorf("failed to delete topolvm controller deployment %s: %w", existingDeployment.GetName(), err)
-	}
-
-	logger.Info("initiated topolvm controller deployment deletion", "TopolvmController", existingDeployment.Name)
 	return nil
 }
 
-func (c topolvmController) setTopolvmControllerDesiredState(existing, desired *appsv1.Deployment) error {
-
+func (c topolvmController) mutateDeployment(existing, desired *appsv1.Deployment) error {
```
suleymanakbas91, line 89: We can remove this function and move its content to the inline update function. This will make it consistent with the rest.

suleymanakbas91 (changes requested):

controllers/vgmanager.go
jakobmoellerdev: that explains why it worked before :D ill remove it from 87 and put it here because then the create/update doesnt need to define the call twice

controllers/topolvm_controller.go
jakobmoellerdev: removed and inlined

PR #405, opened 2023-09-04
OCPBUGS-18354: owns watches config updates

jakobmoellerdev:
This updates all cluster-scoped subresources to have 6 new labels useful for filtering and assignments (in place of the usual ownerreference) - csidriver, volumesnapshotclass,scc,storageclass will now receive these labels. These labels are also used to filter the watch to stop watching for unnecessary objects in the cluster.

On top of this, now LVMVolumeGroupNodeStatus receives OwnerReferences from LVMVolumeGroup. We can later decide to upgrade them to blockOwnerDeletion as well, but currently its a normal OwnerReference (no controller reference to avoid conflicts between multiple volume groups on one NodeStatus). If the volumegroup disappears from the node, so does the owner reference.

Snapshotting was moved to a similar check as SecurityContextConstraints to also allow a watch on the correct snapshot class. Sideeffect: if LVMS is installed after snapshot apis, a manual restart is necessary. For all openshift distros and people aware of this, this shouldnt be an issue since the APIs are usually installed before extension operators.

Also adds an Owns configuration to the LVMCluster for the deployment controller of topo lvm in addition to the already watched daemon set.

Now all resources except for the initially created CSIDriver are watched and reacted to on status.

### DIFF
--- a/controllers/lvmcluster_controller.go
+++ b/controllers/lvmcluster_controller.go
@@ -46,14 +46,12 @@ import (
 type EventReasonInfo string
 type EventReasonError string
 
-const EventReasonErrorDeletionPending EventReasonError = "DeletionPending"
-const EventReasonErrorResourceReconciliationFailed EventReasonError = "ResourceReconciliationFailed"
-const EventReasonResourceReconciliationSuccess EventReasonInfo = "ResourceReconciliationSuccess"
-
-var lvmClusterFinalizer = "lvmcluster.topolvm.io"
-
 const (
-	ControllerName = "lvmcluster-controller"
+	EventReasonErrorDeletionPending              EventReasonError = "DeletionPending"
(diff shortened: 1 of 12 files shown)


controllers/lvmcluster_controller.go
```diff
@@ -46,14 +46,14 @@ import (
 type EventReasonInfo string
 type EventReasonError string
 
-const EventReasonErrorDeletionPending EventReasonError = "DeletionPending"
-const EventReasonErrorResourceReconciliationFailed EventReasonError = "ResourceReconciliationFailed"
-const EventReasonResourceReconciliationSuccess EventReasonInfo = "ResourceReconciliationSuccess"
-
-var lvmClusterFinalizer = "lvmcluster.topolvm.io"
+const (
```
suleymanakbas91, line 49: Could you combine the consts?

PR #410, opened 2023-09-08
OCPVE-677: chore: separate filter and lvm package from vgmanager

jakobmoellerdev:
Part of the findings of vgmanager testing. This separates vgmanager into 3 packages:

1. lvm, now contains ALL calls to LVM in the codebase
2. filter, now contains our filter methods and gets included from vgmanager
3. vgmanager, now only calls to lvm and filter in the correct places

Also the previously raw command executor calls for thinpool creation/extension were moved into separate methods in lvm package so that the perogative to contain all lvm calls still remains.

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -14,13 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package vgmanager
+package filter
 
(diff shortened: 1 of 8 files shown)


suleymanakbas91:
I loved this! 
/lgtm
/approve

PR #413, opened 2023-09-08
OCPVE-220: remove unused certificates

suleymanakbas91:
This PR removes the initcontainer `self-signed-cert-generator` from TopoLVM Controller. This is because the certs are used for TopoLVM webhooks which are not deployed in LVMS. 

Also, this PR removes the unused webhook and certmanager files to simplify the codebase.


config/crd/kustomization.yaml
```diff
@@ -18,21 +18,6 @@ patches:
       name: lvmclusters.lvm.topolvm.io
     path: patches/additionalPrinterColumn_in_lvmcluster.yaml
 
-patchesStrategicMerge:
```
jakobmoellerdev, line 21: I just want to mention here that removing these parts will force us to never work outside an openshift cluster with certificates ever. People usually need cert manager patches outside otherwise the webhooks will not work.

jakobmoellerdev: Clarified: can be removed

controllers/topolvm_controller.go
```diff
@@ -96,16 +98,14 @@ func (c topolvmController) ensureDeleted(r *LVMClusterReconciler, ctx context.Co
...
 	// Topolvm CSI Controller Deployment
 	var replicas int32 = 1
 	volumes := []corev1.Volume{
 		{Name: "socket-dir", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
-		{Name: "certs", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
-	}
-
-	initContainers := []corev1.Container{
-		initContainer(initImage),
+		// TopoLVM needs the certs for webhooks that are not deployed in LVMS, but there is no option to disable them from the controller webhook server.
```
jakobmoellerdev, line 106: Why dont we deploy the webhooks for TopoLVM? Maybe also an oversight on TopoLVM side, if we dont use any webhooks we shouldnt need to mount them right? Maybe worth a PR on TopoLVM?

docs/design/lvm-operator-manager.md
```diff
@@ -33,7 +33,7 @@ The `csiDriver` reconcile unit creates the TopoLVM `CSIDriver` resource.
 
 ### TopoLVM Controller
 
-The `topolvmController` reconcile unit is responsible for deploying a single instance of the TopoLVM controller plugin deployment and ensuring that any necessary updates are made to the deployment. As part of this process, an init container is used to generate openssl certificates that are utilized by the TopoLVM Controller. However, it should be noted that this method will be replaced with the use of cert-manager in the near future.
+The `topolvmController` reconcile unit is responsible for deploying a single instance of the TopoLVM controller plugin deployment and ensuring that any necessary updates are made to the deployment.
```
jakobmoellerdev, line 36: Do we even need this sentence anymore? Kind of feel like it doesn't say anything.

jakobmoellerdev (commented):
Less certs = more good!

Had a question regarding the necessary mount though since Im pretty sure this can be fixed in upstream. They already have flags in their helm chart for disabling it but maybe we also should contribute a code path that allows also omitting the mount. WDYT?

controllers/topolvm_controller.go
suleymanakbas91: Yes, I agree this is worth a PR on TopoLVM. I was going to propose this as an interim solution, but I guess this PR can wait until we improve this part on TopoLVM.

PR #423, opened 2023-09-14
OCPVE-600: Implement the ability to wipe the devices

suleymanakbas91:


pkg/wipefs/wipefs.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/openshift/lvm-operator/pkg/internal/exec"
+)
+
+var (
+	DefaultWipefs        = "/usr/sbin/wipefs"
+	DefaultDMSetup       = "/usr/sbin/dmsetup"
```
jakobmoellerdev, line 12: maybe worth extracting dmsetup into its own interface and separating these two

pkg/wipefs/wipefs.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// Wipe wipes the device only if force delete flag is set
+func (wipefs *HostWipefs) Wipe(deviceName string) error {
+	if len(deviceName) == 0 {
+		return fmt.Errorf("failed to wipe the device. Device name is empty")
+	}
+
+	args := []string{"--all", "--force"}
```
jakobmoellerdev, line 45: Should we try to use non force first, and use force if the policy (see my above comment) allows it? Or is it always mandatory to run force?

pkg/vgmanager/wipe_devices.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (r *VGReconciler) wipeDevicesIfNecessary(ctx context.Context, volumeGroup *lvmv1alpha1.LVMVolumeGroup, vgs []lvm.VolumeGroup, blockDevices []lsblk.BlockDevice) (bool, error) {
+	if volumeGroup.Spec.DeviceSelector == nil || !volumeGroup.Spec.DeviceSelector.ForceWipeDevicesAndDestroyAllData {
+		return false, nil
+	}
+
+	allPaths := append(volumeGroup.Spec.DeviceSelector.Paths, volumeGroup.Spec.DeviceSelector.OptionalPaths...)
+	wiped := false
+	for _, path := range allPaths {
+		if isDeviceAlreadyPartOfVG(vgs, path, volumeGroup) {
```
jakobmoellerdev, line 24: I believe if the device is part of the vg and the vg was already creatd once and is discoverable in the node status, we can just skip this path. However, if the device is part of a vg that is not in the node status, we have to decide if we should risk killing the existing vg that is not managed by us by wiping the disk, or even erroring. WDYT?

suleymanakbas91: I've changed the implementation to check `isDeviceAlreadyPartOfVG` by verifying the existence from the `LVMVolumeGroupNodeStatus` instead of from `vgs` call. This will prevent a case where the name of a user-created volume group is the same as the one created in LVMS, and we assume that it is created by LVMS, and skip these devices.

api/v1alpha1/lvmcluster_types.go
```diff
@@ -114,6 +114,12 @@ type DeviceSelector struct {
 	// We discourage using the device names as they can change over node restarts.
 	// +optional
 	OptionalPaths []string `json:"optionalPaths,omitempty"`
+
+	// ForceWipeDevicesAndDestroyAllData runs wipefs to wipe the devices.
+	// This can lead to data lose. Please use this only when you know that the disk
+	// does not contain any important data.
+	// +optional
+	ForceWipeDevicesAndDestroyAllData bool `json:"forceWipeDisksAndDestroyAllData,omitempty"`
```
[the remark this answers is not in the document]
suleymanakbas91: This was already aligned with the PM and LSO team, and I'd like to keep it as a simple feature and not add more complexity to the codebase. So, I'd like to keep it as is if you don't think it's too bad.

pkg/wipefs/wipefs.go
suleymanakbas91: This was the ask from the beginning, force delete the signatures. Also, see the comment above.

pkg/dmsetup/dmsetup.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewHostDmsetup(executor exec.Executor, dmsetup string) *HostDmsetup {
+	return &HostDmsetup{
+		Executor: executor,
+		dmsetup:  dmsetup,
+	}
+}
+
+// Remove removes the device's reference from the device-mapper
+func (dmsetup *HostDmsetup) Remove(deviceName string) error {
```
jakobmoellerdev, line 36: Nit, unit test missing for this

pkg/vgmanager/devices.go
```diff
@@ -200,11 +200,14 @@ func (r *VGReconciler) getNewDevicesToBeAdded(ctx context.Context, blockDevices
 	return validBlockDevices, nil
 }
 
-func isDeviceAlreadyPartOfVG(vgs []lvm.VolumeGroup, diskName string, volumeGroup *lvmv1alpha1.LVMVolumeGroup) bool {
-	for _, vg := range vgs {
-		if vg.Name == volumeGroup.Name {
-			for _, pv := range vg.PVs {
-				if pv.PvName == diskName {
+func isDeviceAlreadyPartOfVG(nodeStatus *lvmv1alpha1.LVMVolumeGroupNodeStatus, diskName string, volumeGroup *lvmv1alpha1.LVMVolumeGroup) bool {
+	if nodeStatus == nil {
```
jakobmoellerdev, line 204: If node status is nil that means we never checked the vgs right? So is it unsafe to return false here?

pkg/vgmanager/wipe_devices_test.go
```diff
@@ -0,0 +1,174 @@
...
+			if tt.wipeCount > 0 {
+				mockWipefs.EXPECT().Wipe(mock.Anything).Return(nil).Times(tt.wipeCount)
+			}
+			if tt.removeReferenceCount > 0 {
+				mockDmsetup.EXPECT().Remove(mock.Anything).Return(nil).Times(tt.removeReferenceCount)
+			}
+			volumeGroup := &v1alpha1.LVMVolumeGroup{
+				ObjectMeta: metav1.ObjectMeta{Name: "vg1"},
+				Spec: v1alpha1.LVMVolumeGroupSpec{DeviceSelector: &v1alpha1.DeviceSelector{
+					Paths:                             tt.devicePaths,
```
jakobmoellerdev, line 160: These tests are missing optionalDevicePath cases I think. Did we test what happens when you pass a non existing devicePath to wipefs?

api/v1alpha1/lvmcluster_types.go
```diff
@@ -114,15 +111,13 @@ type DeviceSelector struct {
...
-// }
 
-// type LVMConfig struct {
-// 	thinProvision bool `json:"thinProvision,omitempty"`
-// }
+	// ForceWipeDevicesAndDestroyAllData runs wipefs to wipe the devices.
+	// This can lead to data lose. Enable this only when you know that the disk
+	// does not contain any important data.
+	// +optional
+	ForceWipeDevicesAndDestroyAllData bool `json:"forceWipeDevicesAndDestroyAllData,omitempty"`
```
jakobmoellerdev, line 119: Is the flag-name aligned with LSO or machine config operator already?

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -56,6 +56,12 @@ spec:
                           description: DeviceSelector is a set of rules that should
                             match for a device to be included in the LVMCluster
                           properties:
+                            forceWipeDevicesAndDestroyAllData:
```
jakobmoellerdev, line 59: I believe we should make this property immutable after creation, see https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutability-upon-object-creation or we can use the webhook, whatever you like more

pkg/wipefs/wipefs.go
```diff
@@ -0,0 +1,47 @@
...
+type HostWipefs struct {
+	exec.Executor
+	wipefs string
+}
+
+func NewDefaultHostWipefs() *HostWipefs {
+	return NewHostWipefs(&exec.CommandExecutor{}, DefaultWipefs)
+}
+
+func NewHostWipefs(executor exec.Executor, wipefs string) *HostWipefs {
```
jakobmoellerdev, line 26: Nit, unit test missing for this

jakobmoellerdev (changes requested):
Mostly nits, however I think we should make the new field immutable.

pkg/vgmanager/wipe_devices.go
jakobmoellerdev: If we force wipe devices, shouldn't we also force wipe the devices that were created by LVMS and leftover?

pkg/vgmanager/wipe_devices_test.go
suleymanakbas91: In the implementation, I just append optionalDevicePaths to devicePaths: https://github.com/openshift/lvm-operator/pull/423/files#diff-df5f27c889c2546de41e7ad40a6dbad4e9c7933a22593c529fd23977c25b9e54R19 So, I didn't think it'd make a difference to add optionalDevicePath cases to the test. 

If the devicePath is non-existing, wipefs returns a failure like "No such file or directory". We just fail the reconciliation at that point.

jakobmoellerdev: > If the devicePath is non-existing, wipefs returns a failure like "No such file or directory". We just fail the reconciliation at that point.

In a multi-node scenario, it is normal to have certain optionalDevices that are present on one node but not on another (imagine 2 devices with 2 nodes where each node has one of the devices, but not the other). In this case, it would always fail to wipe the devices on all nodes because at least one device would not be present.

pkg/vgmanager/devices.go
suleymanakbas91: This is only a safeguard for the tests. `nodeStatus` is created if not found at the beginning of the reconciliation, so it will never be nil.

jakobmoellerdev (approved):
/lgtm

PR #425, opened 2023-09-18
OCPVE-352: Remove unused metrics-exporter

suleymanakbas91:
This PR removes the metrics-exporter which was only exporting a test metric. 

It also fixes a bug where lvm operator was not exposing controller runtime metrics.


config/prometheus/monitor.yaml
```diff
@@ -17,8 +17,8 @@ spec:
       tlsConfig:
         insecureSkipVerify: true
     - path: /metrics
-      port: lvms-metrics
-      scheme: http
+      port: https
```
jakobmoellerdev, line 20: Is the servicemonitor via tls supported normally or is there any config change we need to do?

jakobmoellerdev (commented):
Awesome removal of old stuff. Just had one question, other than that LGTM

jakobmoellerdev (approved):
/lgtm

PR #426, opened 2023-09-18
OCPVE-665: feat: report filter status into vgstatus

jakobmoellerdev:
This adds the field `.excluded` to the LVMVolumeGroupNodeStatus. This allows not only to introspect all currently active filters on the VolumeGroup's deviceCandidates, but will also contain a list of all devices that are filtered out by the FilterName in case of unavailable devices. 

It simplifies the filter to only return error instead of bool,error as it did not make any difference in the code path. 

It also make sure that if there are available devices, the status is set to Progressing so that the LVMCluster does not jump from Initial to Failed state without showing Progress. 

It also moves the duplicate device selector path check into a webhook for LVMVolumeGroup instead of being checked every reconcile in vgmanager.

To make the tests in vgmanager more stable I switched from envtest to fakeclient in the vgmanager test. That means we now have a much faster test execution and also dont have the issues of panics in the mock causing bail-out in ginkgo

The output of an LVMCluster started on a CRC node for example gets changed to the following:

```yaml
kind: LVMCluster
metadata:
  # ...
  finalizers:
  - lvmcluster.topolvm.io
  name: my-lvmcluster
  namespace: openshift-storage
  # ...
spec:
  storage:
    deviceClasses:
    - default: true
      fstype: xfs
      name: vg1
      thinPoolConfig:
        name: thin-pool-1
        overprovisionRatio: 10
        sizePercent: 90
status:
  deviceClassStatuses:
  - name: vg1
    nodeStatus:
    - devices:
      - /dev/loop0
      excluded:
      - name: /dev/vda
        reasons:
        - /dev/vda has children and should not be considered
      - name: /dev/vda1
        reasons:
        - /dev/vda1 has the invalid "BIOS-BOOT" partition label
      - name: /dev/vda2
        reasons:
        - /dev/vda2 has invalid filesystem signature vfat and cannot be used
      - name: /dev/vda3
        reasons:
        - /dev/vda3 has bind mounts and cannot be used
        - /dev/vda3 has invalid filesystem signature ext4 and cannot be used
        - /dev/vda3 has the invalid "boot" partition label
      - name: /dev/vda4
        reasons:
        - /dev/vda4 has bind mounts and cannot be used
        - /dev/vda4 has invalid filesystem signature xfs and cannot be used
      name: vg1
      node: crc-2zx29-master-0
      status: Ready
  ready: true
  state: Ready

```


jeff-roche:
General first impressions (I still need to dig in more): I don't love some of the formatting. 
- I'm not sure we need the categorized filters if we have the device and the reason, we might be able to just switch this to an array of filtered out devices and their reason? 
- I'm also not sure what the point of the list of filters is in the "registered" section. Seems a bit verbose.
- I'd also potentially change "filtered" to "filteredOut" to be a little more clear.

jakobmoellerdev:
> General first impressions (I still need to dig in more): I don't love some of the formatting.
> 
> * I'm not sure we need the categorized filters if we have the device and the reason, we might be able to just switch this to an array of filtered out devices and their reason?

I can also flatten the array to make it look like this:

```yaml
spec:
  nodeStatus:
  - devices:
    - /dev/loop0
    filtered:
      devices:
      - name: /dev/mapper/vg1-thin--pool--1
        reasons:
        - /dev/mapper/vg1-thin--pool--1 is an lvm partition and cannot be included
          directly as a device
      - name: /dev/mapper/vg1-thin--pool--1_tdata
        reasons:
        - /dev/mapper/vg1-thin--pool--1_tdata has children and should not be considered
        - /dev/mapper/vg1-thin--pool--1_tdata is an lvm partition and cannot be included
          directly as a device
      - name: /dev/mapper/vg1-thin--pool--1_tmeta
        reasons:
        - /dev/mapper/vg1-thin--pool--1_tmeta has children and should not be considered
        - /dev/mapper/vg1-thin--pool--1_tmeta is an lvm partition and cannot be included
          directly as a device
      - name: /dev/vda
        reasons:
        - /dev/vda has children and should not be considered
      - name: /dev/vda1
        reasons:
        - /dev/vda1 has the invalid "BIOS-BOOT" partition label
      - name: /dev/vda2
        reasons:
        - /dev/vda2 has invalid filesystem signature vfat and cannot be used
      - name: /dev/vda3
        reasons:
        - /dev/vda3 has bind mounts and cannot be used
        - /dev/vda3 has invalid filesystem signature ext4 and cannot be used
        - /dev/vda3 has the invalid "boot" partition label
      - name: /dev/vda4
        reasons:
        - /dev/vda4 has bind mounts and cannot be used
        - /dev/vda4 has invalid filesystem signature xfs and cannot be used
    name: vg1
    status: Ready
```

Is this more to your liking? (almost the same as above but changed the output from KName to Name to output the correct lvm partition instead of the device name)

> * I'm also not sure what the point of the list of filters is in the "registered" section. Seems a bit verbose.

Agreed on this, will remove it in the next iteration.

> * I'd also potentially change "filtered" to "filteredOut" to be a little more clear.

Not sure on this, if you have `.status.devices.filtered` then I assume that the adjective filtered describes that it was removed from a set of objects, instead of being included, since I cannot filter things in, I can only filter things out. Open to other opinions though.

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -301,10 +301,35 @@ spec:
                         properties:
                           devices:
                             description: Devices is the list of devices used by the
-                              deviceclass
+                              VG
```
jeff-roche, line 304: nit: Spell out Volume Group here (I try to avoid acronyms where possible)

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -301,10 +301,35 @@ spec:
...
                             items:
                               type: string
                             type: array
+                          excluded:
+                            description: Excluded contains the per node status of
```
jeff-roche, line 309: What about
> Excluded contains the per node status of devices that were identified as candidates but could not be included in the volume group due to filtering

In particular I think we don't need `"It can be consulted in case the device is not picked up as a device for LVM to use"`

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -301,10 +301,35 @@ spec:
...
+                              in case the device is not picked up as a device for
+                              LVM to use.
+                            items:
+                              properties:
+                                name:
+                                  description: Name is the device that was filtered
+                                  type: string
+                                reasons:
+                                  description: Reasons are the human-readable reasons
+                                    why the device was excluded and could not be used
```
jeff-roche, line 321: I would change it to "... why the device was excluded from the volume group"

Currently the sentence is saying the same thing twice (`excluded`, `could not be used`)

bundle/manifests/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -313,7 +338,8 @@ spec:
                               status
                             type: string
                           status:
-                            description: Status is the status of the VG on the node
+                            description: Status tells if the VG was created on the
```
jeff-roche, line 341: nit: vg acronym

bundle/manifests/lvm.topolvm.io_lvmvolumegroupnodestatuses.yaml
```diff
@@ -45,6 +45,27 @@ spec:
                       items:
                         type: string
                       type: array
+                    excluded:
+                      description: Excluded contains the per node status of applied
```
jeff-roche, line 49: Same comments here as lvm cluster

config/crd/bases/lvm.topolvm.io_lvmclusters.yaml
```diff
@@ -297,10 +297,35 @@ spec:
                         properties:
                           devices:
                             description: Devices is the list of devices used by the
-                              deviceclass
+                              VG
                             items:
                               type: string
                             type: array
+                          excluded:
```
jeff-roche, line 304: Same description comments

config/crd/bases/lvm.topolvm.io_lvmvolumegroupnodestatuses.yaml
```diff
@@ -45,6 +45,27 @@ spec:
                       items:
                         type: string
                       type: array
+                    excluded:
```
jeff-roche, line 48: Same description comments

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
+			return nil
 		},
 
-		noBiosBootInPartLabel: func(dev lsblk.BlockDevice) (bool, error) {
-			biosBootInPartLabel := strings.Contains(strings.ToLower(dev.PartLabel), strings.ToLower("bios")) ||
-				strings.Contains(strings.ToLower(dev.PartLabel), strings.ToLower("boot"))
-			return !biosBootInPartLabel, nil
+		notSuspended: func(dev lsblk.BlockDevice) error {
+			if dev.State == StateSuspended {
+				return fmt.Errorf("%s cannot be in suspended", dev.Name)
```
jeff-roche, line 89:
```suggestion
				return fmt.Errorf("%s cannot be in a suspended state", dev.Name)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
+			return nil
 		},
 
-		noReservedInPartLabel: func(dev lsblk.BlockDevice) (bool, error) {
-			reservedInPartLabel := strings.Contains(strings.ToLower(dev.PartLabel), "reserved")
-			return !reservedInPartLabel, nil
+		noInvalidPartitionLabel: func(dev lsblk.BlockDevice) error {
+			for _, invalidLabel := range invalidPartitionLabels {
+				if strings.Contains(strings.ToLower(dev.PartLabel), invalidLabel) {
+					return fmt.Errorf("%s has the invalid %q partition label", dev.Name, dev.PartLabel)
```
jeff-roche, line 97:
```suggestion
					return fmt.Errorf("%s has an invalid partition label '%q'", dev.Name, dev.PartLabel)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
 
 			// if fstype is set to LVM2_member then it already was created as a PV
 			// this means that if the disk has no children, we can safely reuse it if it's a valid LVM PV.
-			if dev.FSType == "LVM2_member" && !dev.HasChildren() {
-				pvs, err := lvm.ListPVs("")
+			if dev.FSType == FSTypeLVM2Member {
+				pvs, err := lvmInstance.ListPVs("")
 				if err != nil {
-					return false, fmt.Errorf("could not determine if block device has valid filesystem signature, since it is flagged as LVM2_member but physical volumes could not be verified: %w", err)
+					return fmt.Errorf("could not determine if %s has valid filesystem signature, since it is flagged as LVM2_member but physical volumes could not be verified: %w", dev.Name, err)
```
jeff-roche, line 114:
```suggestion
					return fmt.Errorf("could not determine if %s has a valid filesystem signature. It is flagged as a LVM2_member but the physical volumes could not be verified: %w", dev.Name, err)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
+					return nil
 				}
 
-				// if there was no PV that matched it and it still is flagged as LVM2_member, it is formatted but not recognized by LVM
-				// configuration. We can assume that in this case, the Volume can be reused by simply recalling the vgcreate command on it
-				return true, nil
+				// 2. Children and has a volume group that matches the one we want to filter for (it is already used)
+				if foundPV.VgName != vg.GetName() {
+					return fmt.Errorf("%s is already LVM2_Member in %s instead of the desired %s and cannot be initialized",
+						dev.Name, foundPV.VgName, vg.GetName())
```
jeff-roche, line 146:
```suggestion
return fmt.Errorf("%s is already a LVM2_Member in another volume group (%s) and cannot be initialized", dev.Name, foundPV.VgName)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
+					// if there was no PV that matched it and it still is flagged as LVM2_member, it is formatted but not recognized by LVM
+					// configuration. We can assume that in this case, the Volume can be reused by simply recalling the vgcreate command on it
+					return nil
+				}
+
+				// a volume is a valid PV if it exists under the same name as the Block Device and has either
+				// 1. No Children, No Volume Group attached to it and Free Capacity (then we can reuse it)
+				if !dev.HasChildren() {
+					if foundPV.VgName != "" {
+						return fmt.Errorf("%s is already part of another volume group %s and cannot be used", dev.Name, foundPV.VgName)
```
jeff-roche, line 135:
```suggestion
						return fmt.Errorf("%s is already part of another volume group (%s) and cannot be used", dev.Name, foundPV.VgName)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
+				// 2. Children and has a volume group that matches the one we want to filter for (it is already used)
+				if foundPV.VgName != vg.GetName() {
+					return fmt.Errorf("%s is already LVM2_Member in %s instead of the desired %s and cannot be initialized",
+						dev.Name, foundPV.VgName, vg.GetName())
+				} else {
+					return fmt.Errorf("%s is already LVM2_Member in %s: %w", dev.Name, vg.GetName(), ErrDeviceAlreadySetupCorrectly)
+				}
 			}
-			return false, nil
+			return fmt.Errorf("%s has invalid filesystem signature %s and cannot be used", dev.Name, dev.FSType)
```
jeff-roche, line 151:
```suggestion
			return fmt.Errorf("%s has an invalid filesystem signature (%s) and cannot be used", dev.Name, dev.FSType)
```

pkg/filter/filter.go
```diff
@@ -38,97 +40,153 @@ const (
...
 		},
 
-		noChildren: func(dev lsblk.BlockDevice) (bool, error) {
+		noChildren: func(dev lsblk.BlockDevice) error {
 			hasChildren := dev.HasChildren()
-			return !hasChildren, nil
+			if hasChildren {
+				return fmt.Errorf("%s has children and should not be considered", dev.Name)
```
jeff-roche, line 157:
```suggestion
				return fmt.Errorf("%s has children block devices and could not be considered", dev.Name)
```

jeff-roche:
Mostly nits and wording/grammar changes. Nice work!

pkg/vgmanager/devices.go
```diff
@@ -71,119 +72,132 @@ func (r *VGReconciler) addDevicesToVG(ctx context.Context, vgs []lvm.VolumeGroup
...
-	}
+type FilteredBlockDevice struct {
+	lsblk.BlockDevice
+	FilterErrors []error
+}
 
-	return r.filterAvailableDevices(ctx, availableDevices), nil
+type FilteredBlockDevices struct {
+	Available []lsblk.BlockDevice
+	Filtered  []FilteredBlockDevice
```
suleymanakbas91, line 82: I think renaming `Filtered` to `Excluded` would make sense here as well. Both to unify naming and also to make it more obvious what the name suggests.

api/v1alpha1/lvmvolumegroup_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (in *LVMVolumeGroup) ValidateUpdate(_ runtime.Object) (warnings admission.Warnings, err error) {
+	lvmvolumegrouplog.Info("validate update", "name", in.Name)
+	if err := in.validateDeviceSelector(); err != nil {
+		return nil, fmt.Errorf(".Spec.DeviceSelector is invalid: %w", err)
+	}
+	return nil, nil
+}
+
+func (in *LVMVolumeGroup) validateDeviceSelector() error {
```
suleymanakbas91, line 39: Why don't we validate this in LVMCluster webhook?

suleymanakbas91 (changes requested):

api/v1alpha1/lvmvolumegroup_webhook.go
jakobmoellerdev: We do actually, just with more detailed errors: `verifyNoDeviceOverlap` in the LVMCluster webhook contains a similar logic to `validateDeviceSelector`. I think we can also decide to remove this validation completely, but then it would be possibly to create an LVMVolumeGroup externally that skips this validation. Alternatively we can also unify the logic of both paths, however then I have to change the error reporting at one of the two places since they report duplicates differently. What's your preference here?

suleymanakbas91: As `LVMVolumeGroup` is an internally used CRD, I'd vote for removing it for simplicity. If someone creates an `LVMVolumeGroup` CR externally, it might lead to bigger problems than duplicate device paths anyway.

PR #432, opened 2023-09-25
OCPVE-662: docs: v2 api proposal

jakobmoellerdev:
Request to merge the rejected v2 API Proposal

### DIFF
--- a/docs/design/proposals/logical-volume-manager-storage-api-v2.md
+++ b/docs/design/proposals/logical-volume-manager-storage-api-v2.md
@@ -1,0 +1,596 @@
+---
+title: logical-volume-manager-storage-api-v2
+authors:
+  - @jakobmoellerdev
+reviewers: # Include a comment about what domain expertise a reviewer is expected to bring and what area of the enhancement you expect them to focus on. For example: - "@networkguru, for networking aspects, please look at IP bootstrapping aspect"
+  - @brandisher, LVM Operator team manager
+  - @suleymanakbas91
+  - @jeff-roche
+approvers: # A single approver is preferred, the role of the approver is to raise important questions, help ensure the enhancement receives reviews from all applicable areas/SMEs, and determine when consensus is achieved such that the EP can move forward to implementation.  Having multiple approvers makes it difficult to determine who is responsible for the actual approval.
+  - @jerpeter1, LVM Operator staff engineer
+  - @suleymanakbas91
+  - @jeff-roche
+api-approvers: # In case of new or modified APIs or API extensions (CRDs, aggregated apiservers, webhooks, finalizers). If there is no API change, use "None"
+  - @suleymanakbas91
+  - @jeff-roche
+creation-date: 2023-09-22
+last-updated: 2023-09-22
+tracking-link: # link to the tracking ticket (for example: Jira Feature or Epic ticket) that corresponds to this enhancement
+  - [OCPVE-662](https://issues.redhat.com/browse/OCPVE-662)
+  - [OCPVE-703](https://issues.redhat.com/browse/OCPVE-703)
+see-also:
+  - N/A
+replaces:
+  - N/A
+superseded-by:
+  - N/A
+---
+
+# Logical Volume Manager Storage API v2
+
+## Summary
+
+This document is an Enhancement Proposal for implementing a new API version for the Logical Volume Manager Storage.
+The objective of the proposal is to eliminate the assumptions made for single-node development, enabling greater configurability and ease of debugging.
+It introduces changes such as the use of a Set and template which are similar to DaemonSet and Deployment respectively from Kubernetes.
+The status aggregation and API interaction have been extensively rewritten.
+Volume groups now can be reused by different device classes.
+A transition from v1alpha1 to v2alpha1 will be provided with detailed guides, and v1alpha1 will be actively maintained until the new API is considered stable.
+Risks involved include possible side effects from the adaptation and potential bugs from the major rewrite, but the authors propose that these can be mitigated through thorough documentation and extended active support of the old API version.
+
+## Motivation
+
+This enhancement proposal presents a comprehensive plan to develop the subsequent version of this API, with the goal of eliminating the limitations originated from assumptions made for single-node development.
+Our aim is to converge our practices towards true best-practice API conventions adhering to upstream standards.
+The proposed changes are designed to facilitate future development by permitting additional non-breaking changes.
+This, in turn, should expedite the iteration process for developers aiming to report extra status updates.
+Furthermore, we aim to simplify the process of debugging and collecting issues when encountering problems with LVMS, which currently lacks a streamlined reporting system. Hence, this proposal ultimately seeks to mitigate such issues and improve the overall user experience.
+
+### User Stories
+
+* As an OpenShift engineer, I want to be able to quickly debug issues with the Logical Volume Manager Storage (LVMS) without having to analyze separate resources for the desired and actual state. This means that the system should clearly display the difference between the expected and current system states, which would allow me to identify and resolve any discrepancies more efficiently.
+* As an OpenShift engineer, I aim to comply with established Kubernetes API Specifications. Adhering to these conventions will simplify the process of onboarding other developers, as they're likely familiar with them, and aligns our company's practices with the wider developer community. This also means I can rely on the well-maintained upstream conventions, avoiding the need to reinvent the wheel.
+* As an OpenShift engineer, I want the flexibility to design non-breaking API changes, allowing a faster iteration period on smaller API changes. This would enable iterative development and a more agile workflow, letting me incrementally build and improve our systems without causing system-wide disruptions.
+* As a user of the LVMS API, I want all pertinent data for Volume Groups and Nodes readily available in one location. Having a centralized place for all related information will improve usability and make the process of managing storage configurations more straightforward and efficient.
+* As a LVMS API user, I want to utilize my preexisting knowledge of Kubernetes APIs to understand the LVMS APIs more quickly. Leveraging knowledge I already have will reduce the learning curve and allow me to work more effectively and efficiently.
+* As an OpenShift cluster administrator, I need to limit the minimum permissions of LVMS to namespace-level as much as possible. By doing so, I can mitigate the risk of privilege escalation into other namespaces, enhancing the overall security of the system. This would help me maintain the integrity and confidentiality of data across different namespaces, providing peace of mind and aiding in regulatory compliance.
+* As an OpenShift cluster administrator, I wish to have the flexibility to deploy multiple configurations of the Logical Volume Manager Storage (LVMS) tailored to different use cases.
+  This implies that the system should provide the capability to setup and manage distinct LVMS configurations concurrently. Such a feature would be beneficial in scenarios where different applications or services have diverse storage needs or performance requirements.
+
+
+### Goals
+
+* Change LVMS APIs to be compliant to the [official Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)
+* Change LVMS APIs to be compliant to [KEP-1623: Standardize Conditions](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1623-standardize-conditions/README.md)
+* Change LVMS APIs to be compliant with [Cluster/Namespaced Operator-Scope definitions](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#overview)
+* Introduce new interfaces for developers to more accurately report statuses to LVMS API users
+
+### Non-Goals
+
+* Determine when and how `v1alpha1` gets deprecated or removed. We will have to supply enough information to allow transitioning to `v2alpha1` but should not immediately get rid of `v1alpha1`.
+
+## Proposal
+
+This is where we get down to the nitty gritty of what the proposal
+actually is. Describe clearly what will be changed, including all of
+the components that need to be modified and how they will be
+different. Include the reason for each choice in the design and
+implementation that is proposed here, and expand on reasons for not
+choosing alternatives in the Alternatives section at the end of the
+document.
+
+### Workflow Description
+
+Explain how the user will use the feature. Be detailed and explicit.
+Describe all of the actors, their roles, and the APIs or interfaces
+involved. Define a starting state and then list the steps that the
+user would need to go through to trigger the feature described in the
+enhancement. Optionally add a
+[mermaid](https://github.com/mermaid-js/mermaid#readme) sequence
+diagram.
+
+Use sub-sections to explain variations, such as for error handling,
+failure recovery, or alternative outcomes.
+
+For example:
+
+**cluster creator** is a human user responsible for deploying a
+cluster.
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,587 @@
...
+This, in turn, should expedite the iteration process for developers aiming to report extra status updates.
+Furthermore, we aim to simplify the process of debugging and collecting issues when encountering problems with LVMS, which currently lacks a streamlined reporting system. Hence, this proposal ultimately seeks to mitigate such issues and improve the overall user experience.
+
+### User Stories
+
+* As an OpenShift engineer, I want to be able to quickly debug issues with the Logical Volume Manager Storage (LVMS) without having to analyze separate resources for the desired and actual state. This means that the system should clearly display the difference between the expected and current system states, which would allow me to identify and resolve any discrepancies more efficiently.
+* As an OpenShift engineer, I aim to comply with established Kubernetes API Specifications. Adhering to these conventions will simplify the process of onboarding other developers, as they're likely familiar with them, and aligns our company's practices with the wider developer community. This also means I can rely on the well-maintained upstream conventions, avoiding the need to reinvent the wheel.
```
jeff-roche, line 52: You might be able to more succinctly say this as

> As an OpenShift engineer, I aim to comply with established Kubernetes API Specifications to align with community and company best practices.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,587 @@
...
+* As an OpenShift engineer, I want the flexibility to design non-breaking API changes, allowing a faster iteration period on smaller API changes. This would enable iterative development and a more agile workflow, letting me incrementally build and improve our systems without causing system-wide disruptions.
+* As a user of the LVMS API, I want all pertinent data for Volume Groups and Nodes readily available in one location. Having a centralized place for all related information will improve usability and make the process of managing storage configurations more straightforward and efficient.
+* As a LVMS API user, I want to utilize my preexisting knowledge of Kubernetes APIs to understand the LVMS APIs more quickly. Leveraging knowledge I already have will reduce the learning curve and allow me to work more effectively and efficiently.
```
jeff-roche, line 55: Suggested change:

> As a LVMS API user, I want to utilize my preexisting knowledge of Kubernetes APIs *to have a base level understanding of the LVMS API*. Leveraging knowledge I already have will reduce the learning curve and allow me to work more effectively and efficiently.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,587 @@
...
+* As an OpenShift cluster administrator, I need to limit the minimum permissions of LVMS to namespace-level as much as possible. By doing so, I can mitigate the risk of privilege escalation into other namespaces, enhancing the overall security of the system. This would help me maintain the integrity and confidentiality of data across different namespaces, providing peace of mind and aiding in regulatory compliance.
```
jeff-roche, line 56: I'm confused what this one has to do with api v2

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,587 @@
...
+* As an OpenShift cluster administrator, I wish to have the flexibility to deploy multiple configurations of the Logical Volume Manager Storage (LVMS) tailored to different use cases.
```
jeff-roche, line 57: I think this can be simplified to:

> As an OpenShift cluster administrator, I wish to have the flexibility to deploy multiple concurrent configurations of the Logical Volume Manager Storage (LVMS) tailored to different use cases.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,587 @@
...
+  This implies that the system should provide the capability to setup and manage distinct LVMS configurations concurrently. Such a feature would be beneficial in scenarios where different applications or services have diverse storage needs or performance requirements.
```
jeff-roche, line 58: With the previous change I think this line can be removed

docs/design/proposals/logical-volume-manager-storage-api-v2.md
jakobmoellerdev: You're right I dont think this is bound to the v2 APIs. I was mainly referring to us probably needing either a fully namespaced or cluster-scoped operator.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,585 @@
...
+11. `LVMNode.status.state` is set to `Ready` after all conditions succeed.
+12. `LVMNodeSetReconciler` picks up the `Ready` condition and sets `LVMNodeSet.Status.Conditions[].(type=LVMNodesReady)` to `True`
+13. `LVMNodeSet.status.state` is set to `Ready` after all conditions succeed.
+
+### API Extensions
+
+#### Summary
+
+1. The usage of a `LVMNodeSet` akin to `DaemonSet`. This is used to align the different `Node`'s into one single configurable object.
+2. The introduction of a `spec.template`, similar to a `Deployment`. This template will be used by all nodes in the `LVMNodeSet` to configure individual nodes.
```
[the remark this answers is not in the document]
jakobmoellerdev: I based this on the de-facto standard of using `.spec.template` as downstream generation, like https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template so it is easy to adopt for people knowing k8s primitives. We can choose to deviate from that, but then it looks less "standard". WDYT?

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,585 @@
...
+                - key: serial
+                  operator: In
+                  values:
+                    - "2313AW459310"
+    deviceClasses:
+      - name: thinly-provisioned
+        strategy: Thin
+        volumeGroup: vg1
+        default: true
+        thin:
```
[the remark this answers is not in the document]
jakobmoellerdev: I believe thin implies that we use a pool of storage for provisioning, because otherwise we would have thick provisioning and I am not aware of other thin provisioning options. Any objections to that? Otherwise I can change it

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,585 @@
...
+          deviceSelectorTerms:
+            - matchExpressions:
+                - key: serial
+                  operator: In
+                  values:
+                    - "2313AW459310"
+    deviceClasses:
+      - name: thinly-provisioned
+        strategy: Thin
+        volumeGroup: vg1
```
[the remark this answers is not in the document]
jakobmoellerdev: > Another thought, is it possible to relate multiple vgs to the same thin pool or is it 1-to-1 only?

1-to-1 like you are correctly assuming: A thinpool must exist on exactly one volume group containing n devices. AFAIK it's not possible to create thin-pools from multiple volume groups.

> Can you expand this section to show how other volume groups would also be realted to thin pools here? It's verbose but I think valuable in this situation

What would be your expectation here? Just some more documentation or another volume group being used by a deviceClass? Because of the 1-to-1 relation, im not sure if there is a lot of variance I can add to this example.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,585 @@
...
+## Proposal
+
+We want to create a breaking change of all APIs involved in LVMS: The LVMCluster, LVMVolumeGroup and LVMVolumeGroupNodeStatus.
+Specifically we want to get rid of the Topology Issues that come from LVMVolumeGroup being responsible for multiple nodes and LVMVolumeGroupNodeStatus being responsible for multiple volume groups.
+
+At the same time, we want to change the resource following `LVMCluster` to no longer be confined to a single resource in the cluster.
+Instead, we want to replace it with a namespaced resource that can be deployed as many times as an administrator desires.
+
+### Workflow Description
+**LVMS Administrator** is a human user responsible for deploying lvms in an OpenShift cluster.
```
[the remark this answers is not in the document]
jakobmoellerdev: probably yes, what would your expectation be here? Not really sure how to write this.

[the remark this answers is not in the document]
jeff-roche: Also not sure :joy: maybe an equivalent diagram of how it the current workflow is, show that and then flow into the proposed workflow

docs/design/proposals/logical-volume-manager-storage-api-v2.md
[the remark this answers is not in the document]
jeff-roche: Nah, this was a nitpick so just stick to the standard :+1:

docs/design/proposals/logical-volume-manager-storage-api-v2.md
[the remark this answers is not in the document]
jeff-roche: Maybe just explicitly call out that it's a 1 to 1 relationship somewhere and that should suffice. Just something to make it clear. I can easily see someone who doesn't fully understand thin pools not really understanding that the relationship is 1 to 1

docs/design/proposals/logical-volume-manager-storage-api-v2.md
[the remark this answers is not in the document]
jakobmoellerdev: I tried around a few times and couldn't really get a good diagram for this comparison so I just added a few words on a direct comparison, mainly differing in the case "what to do when you want to observe the lvm status on a node" now being based on `LVMNode` and not `LVMVolumeGroup` and `LVMVolumeGroupNodeStatus` anymore

docs/design/proposals/logical-volume-manager-storage-api-v2.md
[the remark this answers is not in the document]
jakobmoellerdev: Added an explanation that the 1:1 limitation of deviceClasses to volumeGroups is reflecting the same limit in the lvm2 system and is thus not problematic.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,591 @@
...
+With this proposed change, it is easily possible to recognize the `lvm` subsystem for single nodes by looking at `LVMNode` alone.
+
+### API Extensions
+
+#### Summary
+
+1. The usage of a `LVMNodeSet` akin to `DaemonSet`. This is used to align the different `Node`'s into one single configurable object.
+2. The introduction of a `spec.template`, similar to a `Deployment`. This template will be used by all nodes in the `LVMNodeSet` to configure individual nodes.
+3. Generic Conditions for checking if the LVMNodeSetup for all nodes is ready.
+4. Separation of `deviceClasses` from `volumeGroups`. This is because VolumeGroups in theory can be reused by different deviceClasses
```
suleymanakbas91, line 133: I think the benefits of this separation need more clarification. The only argument here is `VolumeGroups in theory can be reused by different deviceClasses`, but we also say `Important to note is that any single deviceClass can only be mapped to one volumeGroup.` below. Does this mean 
- device classes will be 1:1 matching with volume groups? If so, what would be the other benefits of this separation?
- multiple device classes can use a single volume group?  If so, what would happen when a thin-typed device class and thick-typed one try to use the same volume group, will this be allowed?

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,591 @@
...
+          namingPolicy: Prefixed
+          prefix: "lvms"
+        snapshotClass:
+          namingPolicy: Prefixed
+          prefix: "lvms"
+status:
+  state: Ready
+  nodes:
+    - name: archive-storage-node-in-antarctica-east1
+      state: Ready
```
suleymanakbas91, line 229: Would adding a `message` field here make sense to increase our flexibility?

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,591 @@
...
+if object behind nodeSetRef.nodeSelector does not contain Node targeted by LVMNode
+    return error on create / update
+```
+
+#### Constraints when using multiple LVMNodeSets
+
+We suggest limiting LVMNodeSets to have singular ownership over any given node. That entails that one `LVMNode` is only able to be managed by one `LVMNodeSet`.
+Whenever an LVMNodeSet is created that would target a node that is already being covered by another `LVMNodeSet`, we suggest rejecting it through admission webhook.
+
+We also suggest binding the TopoLVM deployment / daemonset to a single `LVMNodeSet`, thus meaning that 2 `LVMNodeSet`'s would also deploy 2 instances of `TopoLVM` that should not be conflicting with each other.
```
suleymanakbas91, line 403: Do we know if it's possible to run two TopoLVM instances on a cluster? Could we maybe use `nodeSelector` on TopoLVM Daemonsets to run them only on the nodes specified by `LVMNodeSet`, so there will never be two instances running on the same node?

suleymanakbas91 (changes requested):

docs/design/proposals/logical-volume-manager-storage-api-v2.md
jakobmoellerdev: Basically this means that one deviceClass can be mapped to n volume groups, but there can be at most one volume group assigned to the same deviceClass. A thin-typed vg and a thick-typed vg can be used at the same time without issues as long as the thin-pool doesnt occupy 100% of the space of the vg

docs/design/proposals/logical-volume-manager-storage-api-v2.md
jakobmoellerdev: I thought of the same but didn't do this as it's common to only relay the single source of truth to the managing resource and have all details hidden in the other CRD. of course this can be changed later on to add more fields if necessary since its just a mapping.

docs/design/proposals/logical-volume-manager-storage-api-v2.md
```diff
@@ -0,0 +1,609 @@
...
+
+This enhancement proposal presents a comprehensive plan to develop the subsequent version of this API, with the goal of eliminating the limitations originated from assumptions made for single-node development.
+Our aim is to converge our practices towards true best-practice API conventions adhering to upstream standards.
+The proposed changes are designed to facilitate future development by permitting additional non-breaking changes.
+This, in turn, should expedite the iteration process for developers aiming to report extra status updates.
+Furthermore, we aim to simplify the process of debugging and collecting issues when encountering problems with LVMS, which currently lacks a streamlined reporting system. Hence, this proposal ultimately seeks to mitigate such issues and improve the overall user experience.
+
+## Decision
+
+After unanimous decision by the LVMS team on October 17 2022, the proposal was reviewed and postponed by the OpenShift Storage team.
```
suleymanakbas91, line 52:
```suggestion
After a unanimous decision by the LVMS team on October 17, 2023, the proposal was reviewed and postponed.
```

suleymanakbas91 (changes requested):
Small correction :)

PR #433, opened 2023-09-27
OCPVE-390: use specific cipher suites for kube-rbac-proxy

suleymanakbas91:
This PR passes specific cipher suites for kube-rbac-proxy to make it run more securely. 

This also removes `topolvm-metrics` Role and RoleBinding as it is already covered with `lvms-metrics` [Role](https://github.com/openshift/lvm-operator/blob/main/bundle/manifests/lvms-metrics_rbac.authorization.k8s.io_v1_role.yaml) and [RoleBinding](https://github.com/openshift/lvm-operator/blob/main/bundle/manifests/lvms-metrics_rbac.authorization.k8s.io_v1_rolebinding.yaml).


jakobmoellerdev (approved):
/lgtm

PR #434, opened 2023-09-29
OCPVE-346: chore: move all controllers under `internal/controllers`

jakobmoellerdev:
Moves all controllers (including vgmanager) under `internal/controllers` after templating recommendations based on https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v4


docs/troubleshooting.md
```diff
@@ -63,7 +63,7 @@ If you encounter a failure message such as `no available devices found` while in
 $ lsblk --paths --json -o NAME,ROTA,TYPE,SIZE,MODEL,VENDOR,RO,STATE,KNAME,SERIAL,PARTLABEL,FSTYPE
 ```
 
-This prints information about the disks on the host. Review this information to see why a device is not considered available for LVMS utilization. For example, if a device has partlabel `bios` or `reserved`, or if they are suspended or read-only, or if they have children disks or `fstype` set, LVMS considers them unavailable. Check [filter.go](../pkg/vgmanager/filter.go) for the complete list of filters LVMS makes use of.
+This prints information about the disks on the host. Review this information to see why a device is not considered available for LVMS utilization. For example, if a device has partlabel `bios` or `reserved`, or if they are suspended or read-only, or if they have children disks or `fstype` set, LVMS considers them unavailable. Check [filter.go](../pkg/controllers/vgmanager/filter.go) for the complete list of filters LVMS makes use of.
```
suleymanakbas91, line 66: Should be inside `internal`. The link does not work now.

internal/cluster/namespace.go
```diff
@@ -7,9 +7,9 @@ import (
 
 const OperatorNamespaceEnvVar = "POD_NAMESPACE"
 
-// GetOperatorNamespace returns the Namespace the operator should be watching for changes
+// GetOperatorNamespace returns the GetNamespace the operator should be watching for changes
```
suleymanakbas91, line 10: I guess these Namespace -> GetNamespace changes were done unintentionally.

internal/controllers/constants/constants.go
```diff
@@ -25,7 +25,7 @@ const (
 
 	// CSI Controller container and deployment names
 	TopolvmControllerDeploymentName = "topolvm-controller"
-	TopolvmControllerContainerName  = "topolvm-controller"
+	TopolvmControllerContainerName  = TopolvmControllerDeploymentName
```
suleymanakbas91, line 28: We can just delete `TopolvmControllerContainerName` and use `TopolvmControllerDeploymentName` instead.

internal/controllers/lvmcluster/lvmcluster_controller.go
```diff
@@ -25,13 +25,14 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	lvmv1alpha1 "github.com/openshift/lvm-operator/api/v1alpha1"
-	"github.com/openshift/lvm-operator/pkg/cluster"
+	"github.com/openshift/lvm-operator/internal/cluster"
+	"github.com/openshift/lvm-operator/internal/controllers/constants"
+	resource2 "github.com/openshift/lvm-operator/internal/controllers/lvmcluster/resource"
```
suleymanakbas91, line 30: why is there a 2 in the end? :)

suleymanakbas91 (changes requested):

internal/cluster/namespace.go
jakobmoellerdev: yup, good catch... generic refactoring mistake

internal/controllers/lvmcluster/lvmcluster_controller.go
jakobmoellerdev: i wonder that myself :D

PR #442, opened 2023-10-06
OCPVE-710: chore: bump OPERATOR_SDK_VERSION to 1.32.0

jakobmoellerdev:
Also adds a sed modification to remove `createdAt` from the bundle to avoid change detection triggers. Since its a generic annotation and the bundle validate succeeds we can assume this is safe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,7 @@ OPERATOR_VERSION ?= 0.0.1
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.28.0
 
-# TODO: Upgrades > 1.25.3 will lead to createdAt being included in the CSV which currently breaks our change detection. Thus we need to stick to this version until we have a fix.
-OPERATOR_SDK_VERSION ?= 1.25.3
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -254,6 +253,8 @@ bundle: update-mgr-env manifests kustomize operator-sdk rename-csv build-prometh
 		--patch '[{"op": "replace", "path": "/spec/replaces", "value": "$(REPLACES)"}]'
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --package $(BUNDLE_PACKAGE) --version $(OPERATOR_VERSION) $(BUNDLE_METADATA_OPTS) \
 		--extra-service-accounts topolvm-node,vg-manager,topolvm-controller
+	# now we remove the createdAt annotation as otherwise our change detection will say that a the bundle changed during verification
+	sed -i '/createdAt/d' ./bundle/manifests/$(BUNDLE_PACKAGE).clusterserviceversion.yaml
```
suleymanakbas91, line 257: This does not work on MacOS. Could you adapt it like [here](https://github.com/openshift/lvm-operator/blob/main/Makefile#L71)?

suleymanakbas91 (changes requested):

PR #445, opened 2023-10-10
OCPVE-353: chore: operf refactoring

jakobmoellerdev:
performance test refactoring to prepare runnning in CI.

Supports running in either long-observation modes for idle tests against a cluster, or initiates stress tests automatically by creating pods/pvcs in the cluster and observing that window.

Collected aggregates:
99,95,90 quantiles on CPU and Memory over timeframe of either stress period, or of duration given by flag `long-term-observation-window` if `run-stress=false`

Writes a Quantile report as `.toml` in `--output-directory` or current working directory if not set.

```
go run ./test/performance --help
This script retrieves cpu and memory usage metrics for all the workloads created by the logical volume manager storage (lvms).

It creates <instances> of "busybox" pods using PVCs provisioned via the Storage class provided, when in stress test mode, or only collects metrics in idle.

A report is written as "metrics-<start_unix_ts>-<end_unix_ts>.toml" in <output-directory> or current working directory if not set.

Quantile Units in report:
CPU: mcores
Memory: MiB

Usage:
  go run ./test/performance [flags]

Examples:
Stress: go run ./test/performance -t $(oc whoami -t) -s lvms-vg1 -i 64
Idle: go run ./test/performance -t $(oc whoami -t) --run-stress false --long-term-observation-window=5m

Flags:
  -h, --help                                    help for performance
  -i, --instances int                           Number of Workloads/Pvcs (each Workload uses a different PVC) to create in the test (default 4)
  -w, --long-term-observation-window duration   only used when not running stress test, defines observation windows as duration into the past, e.g. 5m means the last 5 minutes (default 5m0s)
  -n, --namespace string                        Namespace where operator is deployed and the PVCs and test pods will be deployed/undeployed (default "openshift-storage")
  -o, --output-directory string                 output directory for the metrics report, working directry by default
  -p, --pattern string                          Pattern used to build the PVCs/Workloads names (default "operf")
  -r, --run-stress                              defines if stress tests should be used, if false uses long-term observation and does not attempt to create stress resources (default true)
  -s, --storage-class string                    Name of the topolvm storage class that will be used in the PVCs (default "lvms-vg1")
  -t, --token string                            authentication token needed to connect with the Openshift cluster.

```


test/performance/collector.go
```diff
@@ -0,0 +1,236 @@
...
+
+			entry.adder(container.Mem, memEntry.Value)
+			entry.adder(c.Total.Mem, memEntry.Value)
+		}
+	}
+
+	if c.nodes > 0 {
+		c.AVGPerLVMNode.Mem.Quantile90 = c.Total.Mem.Quantile90 / float64(c.nodes)
+		c.AVGPerLVMNode.Mem.Quantile95 = c.Total.Mem.Quantile95 / float64(c.nodes)
+		c.AVGPerLVMNode.Mem.Quantile99 = c.Total.Mem.Quantile95 / float64(c.nodes)
```
suleymanakbas91, line 156:
```suggestion
		c.AVGPerLVMNode.Mem.Quantile99 = c.Total.Mem.Quantile99 / float64(c.nodes)
```

test/performance/collector.go
```diff
@@ -0,0 +1,236 @@
...
+		c.AVGPerLVMNode.Cpu.Quantile90 = c.Total.Cpu.Quantile90 / float64(c.nodes)
+		c.AVGPerLVMNode.Cpu.Quantile95 = c.Total.Cpu.Quantile95 / float64(c.nodes)
+		c.AVGPerLVMNode.Cpu.Quantile99 = c.Total.Cpu.Quantile95 / float64(c.nodes)
```
suleymanakbas91, line 159:
```suggestion
		c.AVGPerLVMNode.Cpu.Quantile99 = c.Total.Cpu.Quantile99 / float64(c.nodes)
```

suleymanakbas91 (changes requested):

jakobmoellerdev:
/hold

testing with https://github.com/openshift/release/pull/44279 on a separate branch to verify perf test execution in CI

PR #446, opened 2023-10-10
OCPVE-726: Activate inactive logical volumes

suleymanakbas91:
This PR adds a logic to activate the inactive logical volumes. This is important for the recovery scenario where the leftover logical volumes from a previous installation can be inactive and need to be activated to be used in LVMS.


jakobmoellerdev (approved):
/lgtm
/approve

nice change :)

PR #447, opened 2023-10-10
OCPVE-603: Add lvms tag to volume groups

suleymanakbas91:
This PR adds an `lvms` tag to the volume groups used by LVMS. Also, this implements a temporary logic to add the tag to the existing volume groups, which is useful especially for upgrades.

### DIFF
--- a/cmd/vgmanager/vgmanager.go
+++ b/cmd/vgmanager/vgmanager.go
@@ -17,10 +17,12 @@ limitations under the License.
 package vgmanager
 
 import (
+	"context"
 	"fmt"
 	"os"
 
 	"github.com/go-logr/logr"
+	lvmv1 "github.com/openshift/lvm-operator/api/v1alpha1"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager/dmsetup"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager/filter"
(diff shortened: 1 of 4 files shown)


cmd/vgmanager/vgmanager.go
```diff
@@ -132,3 +151,52 @@ func run(_ *cobra.Command, _ []string, opts *Options) error {
 
 	return nil
 }
+
+// addTagToVGs adds a lvms tag to the existing volume groups. This is a temporary logic that should be removed in v4.16.
+func addTagToVGs(ctx context.Context, c client.Client, lvm *lvm.HostLVM, nodeName string, namespace string, logger logr.Logger) error {
```
jakobmoellerdev, line 156: if this fails, lvms will crash. is this desired behavior or should we simply continue?

jakobmoellerdev, line 156: test missing for this function

jakobmoellerdev, line 156:
```suggestion
func addTagToVGs(ctx context.Context, c client.Client, lvm lvm.LVM, nodeName string, namespace string) error {
   logger := log.FromContext(ctx)
```

jakobmoellerdev (commented):
just some minor questions and missing test, other than that lgtm :)

cmd/vgmanager/vgmanager.go
suleymanakbas91: Good point. We should continue in my opinion.

suleymanakbas91: I was thinking of covering this within QE testing, as this is temporary logic. Do you think we should add an e2e test for this?

suleymanakbas91: I've at least added a unit test.

jakobmoellerdev (approved):
/lgtm

PR #449, opened 2023-10-11
Update OWNERS

qJkee:
Remove qJkee from owners list


jakobmoellerdev:
@qJkee I think this needs a Jira for tracking, then it can be merged. Or we force-merge but bot is rejecting right now

PR #451, opened 2023-10-16
OCPVE-353: chore: operf refactoring

jakobmoellerdev:
performance test refactoring to prepare runnning in CI.

Supports running in either long-observation modes for idle tests against a cluster, or initiates stress tests automatically by creating pods/pvcs in the cluster and observing that window.

Collected aggregates:
99,95,90 quantiles on CPU and Memory over timeframe of either stress period, or of duration given by flag `long-term-observation-window` if `run-stress=false`

Writes a Quantile report as `.toml` in `--output-directory` or current working directory if not set.

```
go run ./test/performance --help
This script retrieves cpu and memory usage metrics for all the workloads created by the logical volume manager storage (lvms).

It creates <instances> of "busybox" pods using PVCs provisioned via the Storage class provided, when in stress test mode, or only collects metrics in idle.

A report is written as "metrics-<start_unix_ts>-<end_unix_ts>.toml" in <output-directory> or current working directory if not set.

Quantile Units in report:
CPU: mcores
Memory: MiB

Usage:
  go run ./test/performance [flags]

Examples:
Stress: go run ./test/performance -t $(oc whoami -t) -s lvms-vg1 -i 64
Idle: go run ./test/performance -t $(oc whoami -t) --run-stress false --long-term-observation-window=5m

Flags:
  -h, --help                                    help for performance
  -i, --instances int                           Number of Workloads/Pvcs (each Workload uses a different PVC) to create in the test (default 4)
  -w, --long-term-observation-window duration   only used when not running stress test, defines observation windows as duration into the past, e.g. 5m means the last 5 minutes (default 5m0s)
  -n, --namespace string                        Namespace where operator is deployed and the PVCs and test pods will be deployed/undeployed (default "openshift-storage")
  -o, --output-directory string                 output directory for the metrics report, working directry by default
  -p, --pattern string                          Pattern used to build the PVCs/Workloads names (default "operf")
  -r, --run-stress                              defines if stress tests should be used, if false uses long-term observation and does not attempt to create stress resources (default true)
  -s, --storage-class string                    Name of the topolvm storage class that will be used in the PVCs (default "lvms-vg1")
  -t, --token string                            authentication token needed to connect with the Openshift cluster.

```

How to test:

1. Create a cluster and login via user/pass (ideally kubeadmin)
2. Run one of the makefile commands, or alternatively create the LVMCluster manually and then run the tool manually.
3. Observe findings created from the .toml files in the working directory.


Makefile
```diff
@@ -307,6 +307,16 @@ e2e: ginkgo ## Build and run e2e tests.
 	cd test/e2e && $(GINKGO) build
 	cd test/e2e && ./e2e.test --lvm-catalog-image=$(CATALOG_IMG) --lvm-subscription-channel=$(SUBSCRIPTION_CHANNEL) --lvm-operator-install=$(LVM_OPERATOR_INSTALL) --lvm-operator-uninstall=$(LVM_OPERATOR_UNINSTALL) --disk-install=$(DISK_INSTALL) -ginkgo.v
 
+performance-stress-test: ## Build and run stress tests. Requires a fully setup LVMS installation. if you receive an error during running because of a missing token it might be because you have not logged in via token authentication but OIDC. you need a token login to run the performance test.
```
suleymanakbas91, line 310: Could you write a small section in the docs about performance testing using these new make targets?

suleymanakbas91 (commented):
Looking great in general! It just needs a bit of documentation.

Makefile
jakobmoellerdev: You're right, this definetly requires docs. I added a section on both idle and stress testing to the CONTRIBUTING.md for how to run these performance tests.

suleymanakbas91:
Great work!
/lgtm

PR #459, opened 2023-10-23
test: lvm package test coverage 100%

jakobmoellerdev:
You shall write tests. And so we increased coverage to 100% for the LVM package

### DIFF
--- a/internal/controllers/vgmanager/lvm/lvm.go
+++ b/internal/controllers/vgmanager/lvm/lvm.go
@@ -20,19 +20,19 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	osexec "os/exec"
 	"strings"
 
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager/exec"
 )
 
-type lvmError string
-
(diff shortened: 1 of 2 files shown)


internal/controllers/vgmanager/lvm/lvm.go
```diff
@@ -249,9 +249,9 @@ func (hlvm *HostLVM) DeleteVG(vg VolumeGroup) error {
 	}
 
 	for _, pv := range vg.PVs {
-		_, err = hlvm.ExecuteCommandWithOutput(lvmDevicesCmd, "--delpvid", pv.UUID)
+		_, err = hlvm.ExecuteCommandWithOutputAsHost(lvmDevicesCmd, "--delpvid", pv.UUID)
```
suleymanakbas91, line 252: Why do we need to run this as host now?

jakobmoellerdev: it was previously not run as host by accident, this command should never be run with less privileges than the others. It accesses the lvmdevices file

PR #480, opened 2023-11-08
OCPVE-760: Save lvmd config to a configmap instead of a file

suleymanakbas91:
This PR changes the way how VG Manager supplies lvmd config to TopoLVM Node. The config file was saved as a file on the host before. Now, VG Manager creates a ConfigMap that is mounted to TopoLVM Node. This way, TopoLVM does not need the `file-checker` init-container anymore, hence it is deleted.

### DIFF
--- a/cmd/vgmanager/vgmanager.go
+++ b/cmd/vgmanager/vgmanager.go
@@ -21,6 +21,7 @@ import (
 	"os"
 
 	"github.com/go-logr/logr"
+	"github.com/openshift/lvm-operator/internal/cluster"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager/dmsetup"
 	"github.com/openshift/lvm-operator/internal/controllers/vgmanager/filter"
@@ -84,14 +85,18 @@ func NewCmd(opts *Options) *cobra.Command {
 func run(cmd *cobra.Command, _ []string, opts *Options) error {
 	lvm := lvm.NewDefaultHostLVM()
 	nodeName := os.Getenv("NODE_NAME")
-	namespace := os.Getenv("POD_NAMESPACE")
+
+	operatorNamespace, err := cluster.GetOperatorNamespace()
+	if err != nil {
+		return fmt.Errorf("unable to get operatorNamespace: %w", err)
+	}
 
 	setupClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{Scheme: opts.Scheme})
 	if err != nil {
 		return fmt.Errorf("unable to initialize setup client for pre-manager startup checks: %w", err)
 	}
 
-	if err := tagging.AddTagToVGs(cmd.Context(), setupClient, lvm, nodeName, namespace); err != nil {
+	if err := tagging.AddTagToVGs(cmd.Context(), setupClient, lvm, nodeName, operatorNamespace); err != nil {
 		opts.SetupLog.Error(err, "failed to tag existing volume groups")
 	}
 
@@ -109,7 +114,7 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 		LeaderElection:         false,
 		Cache: cache.Options{
 			DefaultNamespaces: map[string]cache.Config{
-				os.Getenv("POD_NAMESPACE"): {},
+				operatorNamespace: {},
 			},
 		},
 	})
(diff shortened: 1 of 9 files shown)


internal/controllers/vgmanager/lvmd/lvmd.go
```diff
@@ -16,57 +25,90 @@ type ThinPoolConfig = lvmd.ThinPoolConfig
...
+		return fmt.Errorf("failed to marshal config file: %w", err)
+	}
+
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.LVMDConfigMapName,
+			Namespace: c.Namespace,
+		},
 	}
+	result, err := ctrl.CreateOrUpdate(ctx, c.Client, cm, func() error {
```
jakobmoellerdev, line 85: nit: I think this should also clear the config map data to avoid custom strings being able to be saved here by users

internal/controllers/vgmanager/lvmd/lvmd.go
```diff
@@ -16,57 +25,90 @@ type ThinPoolConfig = lvmd.ThinPoolConfig
...
+		cm.Data = map[string]string{"lvmd.yaml": string(out)}
+		return nil
+	})
 	if err != nil {
-		return fmt.Errorf("failed to save config file %s: %w", c.path, err)
+		return fmt.Errorf("failed to apply ConfigMap %s: %w", cm.GetName(), err)
 	}
+
+	logger.Info("ConfigMap applied to cluster", "operation", result, "name", cm.GetName())
```
jakobmoellerdev, line 93: this will log on every reconcile, is that too much for us?

internal/controllers/vgmanager/lvmd/lvmd.go
```diff
@@ -16,57 +25,90 @@ type ThinPoolConfig = lvmd.ThinPoolConfig
...
+		return nil, fmt.Errorf("failed to unmarshal config file: %w", err)
+	}
+	return config, nil
 }
 
-func (c FileConfig) Save(config *Config) error {
-	out, err := yaml.Marshal(config)
+func (c FileConfig) Save(ctx context.Context, config *Config) error {
+	logger := log.FromContext(ctx)
+	// remove the old config file if it still exists
```
jakobmoellerdev, line 67: Lets add a TODO to mark it for removal in 4.16?

jakobmoellerdev (commented):
Looks great :) 

Just one question: what happens with lvmd when the file / configmap is gone? 

If you interact with config maps in the controller (especially if you do it on every reconciel), I think you should cache them and restrict the cache to the namespace that the operator is deployed in. Even if lvmd works fine still, we probably need to add a watch to namespaced config maps to react and reconcile in case of deletion requests?

Other than that we are good to go I think :rocket:

internal/controllers/vgmanager/lvmd/lvmd.go
suleymanakbas91: It will actually be logged when there is a change in the config. But, still, I removed it. Logging the failure should suffice here.

internal/controllers/vgmanager/lvmd/lvmd.go
suleymanakbas91: Hmm isn't this line already overwriting the config map data or should we do something extra to clear the data?
`cm.Data = map[string]string{"lvmd.yaml": string(out)}`

cmd/vgmanager/vgmanager.go
```diff
@@ -109,8 +115,10 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 		LeaderElection:         false,
 		Cache: cache.Options{
 			DefaultNamespaces: map[string]cache.Config{
-				os.Getenv("POD_NAMESPACE"): {},
+				operatorNamespace: {},
 			},
+			ByObject: map[client.Object]cache.ByObject{
```
jakobmoellerdev, line 120: I dont thing its necessary to include this in order to be cached. You have a watches and that should already include it for the namespace

internal/controllers/vgmanager/controller.go
```diff
@@ -79,9 +82,35 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&lvmv1alpha1.LVMVolumeGroup{}).
 		Owns(&lvmv1alpha1.LVMVolumeGroupNodeStatus{}, builder.MatchEveryOwner, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&corev1.ConfigMap{}, handler.EnqueueRequestsFromMapFunc(r.getObjsInNamespaceForReconcile)).
 		Complete(r)
 }
 
+// getObjsInNamespaceForReconcile reconciles the object anytime the given object is in the same namespace
+// as the available LVMVolumeGroups.
+func (r *Reconciler) getObjsInNamespaceForReconcile(ctx context.Context, obj client.Object) []reconcile.Request {
```
jakobmoellerdev, line 91: I think you have missing tests for that function

jakobmoellerdev (approved):
/lgtm
:100:

PR #483, opened 2023-11-11
fix detection of multiple default device-classes

elproko:
As issue reported in:
https://issues.redhat.com/browse/OCPBUGS-23181

This is the change which will fix the issue.

### DIFF
--- a/vendor/github.com/topolvm/topolvm/lvmd/device_class_manager.go
+++ b/vendor/github.com/topolvm/topolvm/lvmd/device_class_manager.go
@@ -131,7 +131,7 @@ func ValidateDeviceClasses(deviceClasses []*DeviceClass) error {
 			return fmt.Errorf("stripe-size format is \"Size[k|UNIT]\": %s", dc.Name)
 		}
 	}
-	if countDefault != 1 {
+	if countDefault > 1 {
 		return errors.New("should have only one default device-class")
 	}
 	return nil


jakobmoellerdev:
Hi @elproko, first of all thanks for taking the time and attempting to fix your issue.

I have a few things after reading through your comments in the topolvm pull requests.

Your goal as stated by you was:

> to use topolvm on more than single node with only single default device-class.

This is a working use case that can be done as follows:

Assuming you have Nodes A,B,C with
Node A => /dev/sda
Node B => /dev/sdb
Node C => /dev/sdc 

(you can also use more safe /disk/by-id of course)

you will have to specify the following setup for one default device class:
```yaml
kind: LVMCluster
metadata:
  name: my-lvmcluster
spec:
  storage:
    deviceClasses:
    - name: vg1
      default: true
      thinPoolConfig:
        name: thin-pool-1
        sizePercent: 90
        overprovisionRatio: 10
      nodeSelector:
        nodeSelectorTerms:
          - matchExpressions:
              - key: topology.topolvm.io/node
                operator: In
                values:
                  - n1.openshift.fqdn
                  - n2.openshift.fqdn
                  - n3.openshift.fqdn
      deviceSelector:
        optionalPaths:
          - /dev/sda # use by id instead if you want
          - /dev/sdb
          - /dev/sdc
      fstype: xfs
```

What will happen is that every node will look for any of these three devicePaths and if it finds at least one, it will use that for the volume group. This way you can get your one default device class as well as all devices in all three nodes.

Please let me know if that fixes your issue.

elproko:
Thank You for this information.

I did not use optionalPaths (maybe I missed that option) so that is why I tried with multi VGs to avoid "path not found" like errors.

I would like to test optional paths approach. 

At the other hand, having multi VGs gives very convinient ability - to choose where new PVC should be created - just by pointing it with appropriate storage-class.

As usual: I would like to have cookie and eat cookie :)

Is it possible to use optionalPaths, single VG  and have some ability to point specific node for new PVC?

jakobmoellerdev:
> At the other hand, having multi VGs gives very convinient ability - to choose where new PVC should be created - just by pointing it with appropriate storage-class.

I agree, this is the feature we should discuss upstream since this requires different configuration per node and is currently unsupported in upstream.

> Is it possible to use optionalPaths, single VG and have some ability to point specific node for new PVC?
I do not think this is possible, there hasn't been a customer request for this thus far. I believe we could introduce a hint on the PVC request to override but the point of Dynamic Allocation in Kubernetes is that you don't have to worry about this during provisioning. This is also the biggest advantage of TopoLVM overall because unlike most other local storage providers, it is a fully compliant CSI Driver with Dynamic Storage Provisioning, so I recommend you use either static provisioning with LSO or hostpath support or rely on TopoLVMs distribution / scheduling algorithm to correctly assign storage across nodes.

jakobmoellerdev:
closing this pr due to this being tackled in upstream and reopened

PR #530, opened 2023-12-13
OCPEDGE-549: Remove temporary tagging logic

suleymanakbas91:
This PR removes the temporary tagging logic we introduced in 4.15 to make upgrades from 4.14 to 4.15 seamless. This is no longer needed in 4.16 or later releases.


suleymanakbas91:
Closing as we decided to remove the temporary tagging logic in 4.17 instead.

PR #533, opened 2023-12-14
OCPEDGE-746: feat!: Micro LVMS - Edge / Low Resource Optimized LVMS

jakobmoellerdev:


Dockerfile
```diff
@@ -23,12 +23,21 @@ COPY internal/ internal/
 # Build
 RUN GOOS=$TARGETOS GOARCH=$TARGETARCH go build -mod=vendor --ldflags "-s -w" -a -o lvms cmd/main.go
 
-# vgmanager needs 'nsenter' and other basic linux utils to correctly function
-FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9/ubi-minimal:9.2
+FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9/ubi-minimal:9.3
+
+RUN curl https://mirror.stream.centos.org/9-stream/BaseOS/$(arch)/os/Packages/centos-gpg-keys-9.0-23.el9.noarch.rpm > centos-gpg-keys-9.0-23.el9.noarch.rpm && \
```
suleymanakbas91, line 28: Shall we add a comment about why we need these rpms?

internal/controllers/lvmcluster/resource/vgmanager_daemonset.go
```diff
@@ -234,3 +353,11 @@ func newVGManagerDaemonset(lvmCluster *lvmv1alpha1.LVMCluster, namespace, vgImag
 	setDaemonsetNodeSelector(nodeSelector, &ds)
 	return ds
 }
+
+func getAbsoluteKubeletPath(name string) string {
```
suleymanakbas91, line 357: Duplicate of https://github.com/openshift/lvm-operator/pull/533/files#diff-ac1e4637145c504870edce40ad17e3f81d5b5d4d5e40fc17b47b1dbe01c212e9R334.

internal/controllers/vgmanager/dmsetup/dmsetup.go
```diff
@@ -43,11 +44,15 @@ func (dmsetup *HostDmsetup) Remove(deviceName string) error {
 	args = append(args, deviceName)
 	output, err := dmsetup.ExecuteCommandWithOutputAsHost(dmsetup.dmsetup, args...)
 	if err != nil {
-		if strings.Contains(output, "not found") {
-			return ErrReferenceNotFound
-		}
 		return fmt.Errorf("failed to remove the reference from device-mapper %q: %v", deviceName, err)
 	}
+	data, err := io.ReadAll(output)
```
suleymanakbas91, line 49: Should this be inside the err check like the deleted part?

internal/controllers/vgmanager/lsblk/lsblk.go
```diff
@@ -125,26 +136,72 @@ func (lsblk *HostLSBLK) IsUsableLoopDev(b BlockDevice) (bool, error) {
...
 		}
 	}
+	return flattened
+}
+
+func (lsblk *HostLSBLK) BlockDeviceInfos(bs []BlockDevice) (BlockDeviceInfos, error) {
+	flattenedMap := flattenedBlockDevices(bs)
+
+	blockDeviceInfos := make(BlockDeviceInfos)
+	// file, err := os.Open(lsblk.mountInfo)
```
suleymanakbas91, line 163: I guess these are just some leftover code that can be removed

cmd/vgmanager/vgmanager.go
```diff
@@ -122,6 +144,73 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
...
+		if err != nil {
+			return fmt.Errorf("could not setup topolvm node server: %w", err)
+		}
+		csi.RegisterNodeServer(grpcServer, nodeServer)
+		err = mgr.Add(runners.NewGRPCRunner(grpcServer, constants.DefaultCSISocket, false))
+		if err != nil {
+			return fmt.Errorf("could not add grpc runner for node server: %w", err)
+		}
+
+		if err := mgr.AddReadyzCheck("readyz", func(req *http.Request) error {
```
suleymanakbas91, line 199: Should we get rid of this one as well?

suleymanakbas91 (changes requested):

Dockerfile
jakobmoellerdev: Agreed. Tried to remove the centos repos but couldnt find a way to make that run outside of RHEL systems with a valid subscription (e.g. on a Fedora workstation).

Added a comment:

```
# We use CentOS Stream 9 as our source for e2fsprogs here so that we can offer a fully open source version for development here.
# This allows users without Red Hat Subscriptions (e.g. on a Fedora Workstation) to build and test LVMS.
# Note that we do NOT provide Support for any images built from this Dockerfile. The authoritative source for LVMS is the
# official Red Hat Container Catalog at https://catalog.redhat.com/software/containers/search?gs&q=lvms4%20operator
# which are built with Red Hat builds of the e2fsprogs package which is only available via Red Hat Subscription.
```

internal/controllers/vgmanager/dmsetup/dmsetup.go
jakobmoellerdev: Not sure what you mean by that, could explain?

internal/controllers/vgmanager/lsblk/lsblk.go
jakobmoellerdev: We dont do bind mount checks anymore in microlvms. Thats important to notice if we run into an issue later. Removed the code for now.

PR #571, opened 2024-02-07
OCPBUGS-29188: fix: resolve PV name via Symlink before comparing to KName

jakobmoellerdev:
This allows us to properly exclude and filter symlinked devices that would previously be compared via kname without being resolved. Now we compare them symlinked which results in proper filter results

Added precommit timeout and incremented golang ci lint version


internal/controllers/vgmanager/devices.go
```diff
@@ -206,7 +206,7 @@ func isDeviceAlreadyPartOfVG(nodeStatus *lvmv1alpha1.LVMVolumeGroupNodeStatus, d
 	for _, vgStatus := range nodeStatus.Spec.LVMVGStatus {
 		if vgStatus.Name == volumeGroup.Name {
 			for _, pv := range vgStatus.Devices {
-				if pv == diskName {
+				if resolvedPV, _ := evalSymlinks(pv); resolvedPV == diskName {
```
jakobmoellerdev, line 209: Note to reviewer: the error here can be ignored as its the same as returning false

PR #599, opened 2024-03-28
[WIP] OCPEDGE-927: Introduce Static discovery of devices that allows immutable storage configurations

jakobmoellerdev:
`DeviceDiscoveryPolicy` is a flag to indicate whether the device should be discovered at install time or at runtime (static or dynamic configuration of devices):
- If set to `DeviceDiscoveryPolicyInstallStatic`, the devices will not be added to the VG if they are not present at setup time. 
- If set to `DeviceDiscoveryPolicyRuntimeDynamic`, the devices will be added to the VG if they are present at runtime. By default, the value is set to `RuntimeDynamic`. 
- This field cannot be updated once the LVMCluster is created.

All existing LVMClusters will work in Dynamic Mode and will not be able to be reconfigured to Static Mode. Any new LVMClusters will be able to be changed to Static Mode.

A new LVMCluster with 2 devices (/dev/loop0 available when LVMCluster was created and /dev/loop1 added afterwards) would look like this:

```yaml
kind: LVMCluster
metadata:
  finalizers:
  - lvmcluster.topolvm.io
  generation: 1
  name: my-lvmcluster
  namespace: openshift-storage
spec:
  storage:
    deviceClasses:
    - default: true
      deviceDiscoveryPolicy: InstallStatic
      fstype: xfs
      name: vg1
      thinPoolConfig:
        name: thin-pool-1
        overprovisionRatio: 10
        sizePercent: 90
status:
  deviceClassStatuses:
  - name: vg1
    nodeStatus:
    - deviceDiscoveryPolicy: InstallStatic
      devices:
      - /dev/loop0
      excluded:
      - name: /dev/loop1
        reasons:
        - /dev/loop1 was not part of vg1 at creation (static device discovery enabled)
      - name: /dev/vda
        reasons:
        - /dev/vda has children block devices and could not be considered
        - /dev/vda was not part of vg1 at creation (static device discovery enabled)
      - name: /dev/vda1
        reasons:
        - /dev/vda1 has an invalid partition label "BIOS-BOOT"
        - /dev/vda1 was not part of vg1 at creation (static device discovery enabled)
      - name: /dev/vda2
        reasons:
        - /dev/vda2 has an invalid filesystem signature (vfat) and cannot be used
        - /dev/vda2 was not part of vg1 at creation (static device discovery enabled)
      - name: /dev/vda3
        reasons:
        - /dev/vda3 has an invalid filesystem signature (ext4) and cannot be used
        - /dev/vda3 has an invalid partition label "boot"
        - /dev/vda3 was not part of vg1 at creation (static device discovery enabled)
        reasons:
        - /dev/vda4 has an invalid filesystem signature (xfs) and cannot be used
        - /dev/vda4 was not part of vg1 at creation (static device discovery enabled)
      name: vg1
      node: crc-vlf7c-master-0
      status: Ready
  ready: true
  state: Ready

```

We can later on think about introudcing switching between RuntimeDynamic and InstallStatic (in case we want people to be able to switch between dynamic allocation and "pinning" their configuration) but this has been out of scope for this PR.

### DIFF
--- a/api/v1alpha1/lvmcluster_types.go
+++ b/api/v1alpha1/lvmcluster_types.go
@@ -79,7 +79,7 @@ type DeviceClass struct {
 
 	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
 	// at install time or at runtime (static or dynamic configuration of devices)
-	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at LVMCluster creation time.
 	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
 	// By default, the value is set to RuntimeDynamic.
 	// This field cannot be updated once the LVMCluster is created.
--- a/api/v1alpha1/lvmcluster_types.go
+++ b/api/v1alpha1/lvmcluster_types.go
@@ -61,6 +61,13 @@ const (
 	FilesystemTypeXFS  DeviceFilesystemType = "xfs"
 )
 
+type DeviceDiscoveryPolicy string
+
+const (
+	DeviceDiscoveryPolicyInstallStatic  DeviceDiscoveryPolicy = "InstallStatic"
+	DeviceDiscoveryPolicyRuntimeDynamic DeviceDiscoveryPolicy = "RuntimeDynamic"
+)
+
 type DeviceClass struct {
 	// Name of the class, the VG and possibly the storageclass.
 	// Validations to confirm that this field can be used as metadata.name field in storageclass
@@ -69,6 +76,17 @@ type DeviceClass struct {
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
 	Name string `json:"name,omitempty"`
+
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
+	// By default, the value is set to RuntimeDynamic.
+	// This field cannot be updated once the LVMCluster is created.
+	// +kubebuilder:validation:Enum=InstallStatic;RuntimeDynamic
+	// +kubebuilder:default=RuntimeDynamic
+	// +kubebuilder:validation:Required
+	DeviceDiscoveryPolicy DeviceDiscoveryPolicy `json:"deviceDiscoveryPolicy,omitempty"`
 
 	// DeviceSelector is a set of rules that should match for a device to be included in the LVMCluster
 	// +optional
--- a/api/v1alpha1/lvmvolumegroup_types.go
+++ b/api/v1alpha1/lvmvolumegroup_types.go
@@ -27,7 +27,7 @@ import (
 type LVMVolumeGroupSpec struct {
 	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
 	// at install time or at runtime (static or dynamic configuration of devices)
-	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at LVMCluster creation time.
 	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
 	// By default, the value is set to RuntimeDynamic.
 	// This field cannot be updated once the LVMCluster is created.
--- a/api/v1alpha1/lvmvolumegroup_types.go
+++ b/api/v1alpha1/lvmvolumegroup_types.go
@@ -25,6 +25,17 @@ import (
 
 // LVMVolumeGroupSpec defines the desired state of LVMVolumeGroup
 type LVMVolumeGroupSpec struct {
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
+	// By default, the value is set to RuntimeDynamic.
+	// This field cannot be updated once the LVMCluster is created.
+	// +kubebuilder:validation:Enum=InstallStatic;RuntimeDynamic
+	// +kubebuilder:default=RuntimeDynamic
+	// +kubebuilder:validation:Required
+	DeviceDiscoveryPolicy DeviceDiscoveryPolicy `json:"deviceDiscoveryPolicy,omitempty"`
+
 	// DeviceSelector is a set of rules that should match for a device to be included in this TopoLVMCluster
 	// +optional
 	DeviceSelector *DeviceSelector `json:"deviceSelector,omitempty"`
--- a/api/v1alpha1/lvmvolumegroupnodestatus_types.go
+++ b/api/v1alpha1/lvmvolumegroupnodestatus_types.go
@@ -56,7 +56,7 @@ type VGStatus struct {
 
 	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
 	// at install time or at runtime (static or dynamic configuration of devices)
-	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at LVMCluster creation time.
 	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
 	// By default, the value is set to RuntimeDynamic.
 	// This field cannot be updated once the LVMCluster is created.
--- a/api/v1alpha1/lvmvolumegroupnodestatus_types.go
+++ b/api/v1alpha1/lvmvolumegroupnodestatus_types.go
@@ -53,6 +53,17 @@ type VGStatus struct {
 	// Excluded contains the per node status of applied device exclusions that were picked up via selector,
 	// but were not used for other reasons.
 	Excluded []ExcludedDevice `json:"excluded,omitempty"`
+
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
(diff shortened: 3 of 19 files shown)


api/v1alpha1/lvmcluster_types.go
```diff
@@ -70,6 +77,17 @@ type DeviceClass struct {
 	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
 	Name string `json:"name,omitempty"`
 
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
+	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
+	// By default, the value is set to RuntimeDynamic.
+	// This field cannot be updated once the LVMCluster is created.
+	// +kubebuilder:validation:Enum=InstallStatic;RuntimeDynamic
```
brandisher, line 86: I thought we had something in the OpenShift Conventions about Enums but as it turns out, it was some of my prior art that triggered the thought :smile: 

Drawing on a distant memory, the OpenShift API has a concept of an intentional "empty" or null choice for Enums so a cluster admin can choose to set a value, and then unset it to allow whatever the default is take effect. The main use case as I recall is upgrades where a cluster admin may want to take advantage of a new way of configuring a field or no longer wants it configured. Is that something that'd be valuable here?

Here are some references:
* https://github.com/openshift/enhancements/blob/master/enhancements/dns/forward-dns-over-tls.md
* https://github.com/openshift/api/pull/1110/files#diff-46e86efb5c123e387737010ac03828fda0c5edfa402c432480485d928827236cR135

api/v1alpha1/lvmvolumegroup_types.go
```diff
@@ -25,6 +25,17 @@ import (
 
 // LVMVolumeGroupSpec defines the desired state of LVMVolumeGroup
 type LVMVolumeGroupSpec struct {
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
```
brandisher, line 30:
```suggestion
	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at LVMCluster install time.
```

api/v1alpha1/lvmvolumegroup_types.go
```diff
@@ -25,6 +25,17 @@ import (
...
+	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
+	// By default, the value is set to RuntimeDynamic.
+	// This field cannot be updated once the LVMCluster is created.
+	// +kubebuilder:validation:Enum=InstallStatic;RuntimeDynamic
```
brandisher, line 34: Same question here about the empty/null choice for Enums.

api/v1alpha1/lvmvolumegroupnodestatus_types.go
```diff
@@ -53,6 +53,17 @@ type VGStatus struct {
 	// Excluded contains the per node status of applied device exclusions that were picked up via selector,
 	// but were not used for other reasons.
 	Excluded []ExcludedDevice `json:"excluded,omitempty"`
+
+	// DeviceDiscoveryPolicy is a flag to indicate whether the device should be discovered
+	// at install time or at runtime (static or dynamic configuration of devices)
+	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at setup time.
```
brandisher, line 59:
```suggestion
	// If set to DeviceDiscoveryPolicyInstallStatic, the devices will not be added to the VG if they are not present at LVMCluster install time.
```

api/v1alpha1/lvmvolumegroupnodestatus_types.go
```diff
@@ -53,6 +53,17 @@ type VGStatus struct {
...
+	// If set to DeviceDiscoveryPolicyRuntimeDynamic, the devices will be added to the VG if they are present at runtime.
+	// By default, the value is set to RuntimeDynamic.
+	// This field cannot be updated once the LVMCluster is created.
+	// +kubebuilder:validation:Enum=InstallStatic;RuntimeDynamic
```
brandisher, line 63: Same question here about the empty/null choice for Enums.

internal/controllers/vgmanager/filter/filter_test.go
```diff
@@ -23,14 +24,111 @@ type advancedFilterTestCase struct {
 	lvmExpect []lvm.PhysicalVolume
 }
 
+func TestNewDeviceInStaticVolumeGroup(t *testing.T) {
```
brandisher, line 27: How do we test the y-stream upgrade path where we go from the field not existing, to existing and not being set to a value?

brandisher (commented):
I like this idea a lot - it solves a persistent category of problems that's often unexpected for LVMS users. There are a few inline questions but nothing blocking that I can see.

api/v1alpha1/lvmcluster_types.go
jakobmoellerdev: TBH thats the first time I've heard about this so let me try to understand what youre trying to achieve:

Assuming the LVMCluster is configured
1. The fields would be set to some value
2. The administrator would want to use a new default, and set it to ""
3. The administrator would rely on OpenShift to figure out the new default.

Is that correct? If so im not sure if its applicable, because the Field is immutable after creation of the LVMCluster, meaning changing the value is impossible (at least as its implemented right now). Otherwise we would have to test switching from one setup to the other.

Maybe im missing things though. Either way we can also implement a defaulting logic like you suggested I think. Defaulting to @suleymanakbas91 opinion to break the tie here.

internal/controllers/vgmanager/filter/filter_test.go
jakobmoellerdev: So not being set to a value is not an option due to 
```
	// +kubebuilder:default=RuntimeDynamic
	// +kubebuilder:validation:Required
```
This will force the object to automatically contain RuntimeDynamic if no value is set. Similarly we can assume that the field will be present even for clusters that didnt have it set originally.

That said, the y-stream path would be to do a regular regression run where the LVMCluster gets created with a version that does not have this field change, then updating to the new CRD version, and then running a regression test. There should be no change in behavior. On the next edit the storage version would automatically get the field written in during serialization. I think this is pretty standard K8s behavior.

api/v1alpha1/lvmcluster_types.go
jakobmoellerdev: Ah one more thing, if in a future release the default should change, we would change the CRD annotation accordingly, I don't think there is a mandatory need for a zero-value here.

brandisher: +1 to having @suleymanakbas91 tie break although now that I think about it, I guess a better question from me would be, is there a use case for a mutable value?

After relooking at my reference I realized the the field was optional, but that's not the case here so it might not even apply. The only thing I could think of which is probably a weak justification is "useability" in that the field could be hidden if it's not set, giving users less configuration to key in on. Perhaps a GitOps impacting use case? What do you think?

P.S. I'll resolve the other comments around this since whatever gets decided will have to affect the references anyway.

jakobmoellerdev: On the GitOps part: When applying via e.g. ArgoCD, it will use either SSA or Strategic Merge Patch, in which case fields that are not set in the Argo Specification but are defaulted in through the operator, are ignored by Argo. Flux should behave the same / similar.

jakobmoellerdev: For the mutable value: The use case would be that a user deploys with an initially static configuration, but then has the need to expand the storage later on with some devices he does not know the serials from. I think this is too much of an edge case to consider for the implementation now, so thats why I left it as immutable

jakobmoellerdev:
This depends on how the disk was added to LVMCluster, e.g. which path/symlink is used.
If using a device path such as /dev/xda, then one would need to make sure the device appears under this path again.

You are right in that it is impossible to switch to a new device path like this, however removing device Paths is already not supported today from what I know. Meaning once you enter a device path it cannot be removed (maybe a design / product oversight?)

When using the usual symlinks we have to differentiate:

- /dev/disk/by-path: This is okay to be used as long as the replaced device is slotted into the same PCIe Slot and the Host OS has persistent PCIe Controller mapping
- /dev/disk/by-id: The Disk ID has to match the hardware serial number, this is impossible to replace because there is no way the new disk can match to the old disk

the by-path will already change when the device is plugged into a different port of the controller, the by-id will change when the device is plugged into a port of a hardware controller subject to another subsystem. For example, The /dev/disk/by-path/pci-* paths for NVMe devices may change if a PCIe device is added or removed from the system or if the system firmware [does not enumerate them deterministically](https://github.com/systemd/systemd/issues/22692)

- /dev/disk/by-label: to be preferred, the Disk has to use the same label as the broken disk and will be picked up again in its stead, but the old disk has to be unmounted
- /dev/disk/by-partlabel: equivalent to by-label, but to be used when using a prepartitioned layout with GPT entries.
- /dev/disk/by-uuid: We should not use this one as it uses filesystem UUIDs and we potentially have only block storage to rely on
- /dev/disk/by-partuuid: We should not reference this one as it uses GPT partition UUIDs or MBR pseudo UUIDs and thus make them irreplaceable device paths.

api/v1alpha1/lvmcluster_types.go
brandisher: Restating to make sure I understand; we're saying that it's an edge use case for a cluster admin to go from static configuration with a known set of disks to dynamic with an unknown set of disks?

jakobmoellerdev: Yeah that pretty much sums it up. I think for people who want to try out LVMS they usually start with dynamic configuration, only to later find out that they want to have a static device discovery

brandisher: Ah okay, that makes sense. I'm comfortable resolving this thread with the understanding that I'm unsure if practically there's anything we need to future guard against or plan for -- the rationale for the implementation is clear.

DanielFroehlich:
> This depends on how the disk was added to LVMCluster, e.g. which path/symlink is used. 
I was under the impression that the `DeviceDiscoveryPolicy` is only applied when no `devicePath` is used?

> 
> You are right in that it is impossible to switch to a new device path like this, however removing device Paths is already not supported today from what I know. Meaning once you enter a device path it cannot be removed (maybe a design / product oversight?)

We should make sure this is documented.

> When using the usual symlinks we have to differentiate:
Thanks for the excellent enumeration and explanation. We should add this to the docs.

Overall: This needs a FEATURE, as it also needs docs. We should use the opprtunity to add a "Guide to use disk devices with LVMS". (e.g. recommend HW RAID, make sure the restrictions on device removal are clearly stated,  etc). (I will create that feature)

jakobmoellerdev:
I think those are good Action Items. Also started looking into https://issues.redhat.com/browse/OCPEDGE-1007 based on your previous comment now to get clear understanding from engineering side. We can use that to fill the docs and give us a better idea on what we need to fix.

DanielFroehlich:
> This depends on how the disk was added to LVMCluster, e.g. which path/symlink is used.


I was under the impression that the DeviceDiscoveryPolicy is only applied when no devicePath is used?

jakobmoellerdev:
/hold
putting on hold until we have some more clarity on the approach

jakobmoellerdev:
Closing until we have a way forward since it doesnt seem likely this goes in

PR #602, opened 2024-04-04
chore: update feature annotations for lvms

jakobmoellerdev:


bundle/manifests/lvms-operator.clusterserviceversion.yaml
```diff
@@ -31,6 +31,16 @@ metadata:
     containerImage: quay.io/ocs-dev/lvm-operator:latest
     description: Logical volume manager storage provides dynamically provisioned local
       storage for container workloads
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "true"
+    features.operators.openshift.io/disconnected: "true"
```
suleymanakbas91, line 37: This should be set to `false` as disconnected support was introduced in 4.13.

suleymanakbas91 (changes requested):

PR #612, opened 2024-04-19
OCPBUGS-30830: Accommodate the case in readiness calculation where readyVGCount is greater than expectedVGCount

suleymanakbas91:
This PR updates the readiness calculation logic to accommodate the case where readyVGCount is greater than expectedVGCount, e.g. during an upgrade.


jakobmoellerdev (commented):
As I understand we are still missing one part here from the original triage:

> Additionally, because we dont watch for Node Taints on LVMCluster, the reconcile loop is not refreshed which means one has to either restart lvm-operator or wait until the cache resets (2 hours).

Do we already get the node updates?

jakobmoellerdev (approved):
/lgtm

PR #615, opened 2024-04-24
OCPEDGE-922: Bring more visibility to dynamic device discovery by warnings and a new status field

suleymanakbas91:
This PR adds some admission warnings and more visibility through a status field in `LVMCluster` CRD in case the dynamic device discovery is enabled by not setting a `DeviceSelector`. 

The new status field is called `DeviceDiscoveryPolicy`:
- If set to `DeviceDiscoveryPolicyPreconfigured`, the devices are preconfigured through a `DeviceSelector`. 
- If set to `DeviceDiscoveryPolicyRuntimeDynamic`, the devices are added to the VG if they are present at runtime.


api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -298,23 +300,25 @@ func (v *lvmClusterValidator) verifyDeviceClass(l *LVMCluster) (admission.Warnin
...
 		if deviceClass.DeviceSelector != nil {
 			if len(deviceClass.DeviceSelector.Paths) == 0 && len(deviceClass.DeviceSelector.OptionalPaths) == 0 {
-				return ErrPathsOrOptionalPathsMandatoryWithNonNilDeviceSelector
+				return nil, ErrPathsOrOptionalPathsMandatoryWithNonNilDeviceSelector
 			}
 		} else {
 			deviceClassesWithoutPaths = append(deviceClassesWithoutPaths, deviceClass.Name)
 		}
 	}
 	if len(l.Spec.Storage.DeviceClasses) > 1 && len(deviceClassesWithoutPaths) > 0 {
```
jakobmoellerdev, line 315: I would suggest to change this logic so that we use the for loop above to add an admission warning for every deviceClass with a nil device selector. that way we dont need such a complicated if statement here. WDYT?

api/v1alpha1/lvmvolumegroupnodestatus_types.go
```diff
@@ -22,6 +22,13 @@ import (
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type DeviceDiscoveryPolicy string
+
+const (
+	DeviceDiscoveryPolicyPreConfigured  DeviceDiscoveryPolicy = "PreConfigured"
```
jakobmoellerdev, line 28: These constants are lacking documentation when not navigating from DeviceDiscoveryPolicy docs, maybbe move the docs from DeviceDiscoveryPolicy and write something more here?

jakobmoellerdev: I think Preconfigured is one word, so you dont need to capitalise configured.

jakobmoellerdev (commented):
Nice addition and not as crazy as my previous attempt. Overall lgtm, just a few minor things for docs and a small nit on the warning generation

api/v1alpha1/lvmcluster_webhook.go
suleymanakbas91: The thing is that this warning is valid only for the single device class scenario. For the multi device class, we already return an error if they do not specify a device selector in L#316.

jakobmoellerdev (approved):
/lgtm

PR #616, opened 2024-04-25
OCPEDGE-1005: must-gather: add support for --since and --since-time flags

suleymanakbas91:
`oc adm must-gather` has new flags `--since` and `--since-time`, which set `MUST_GATHER_SINCE` and `MUST_GATHER_SINCE_TIME` environment vars to limit log collection based on a specified time. This PR adds support for these flags to LVMS must-gather.


jakobmoellerdev (approved):
/lgtm

PR #617, opened 2024-04-30
OCPBUGS-32954: OCPBUGS-32753: Remove LVMVolumeGroupNodeStatus CRs when LVMCluster CR is removed

suleymanakbas91:


internal/controllers/lvmcluster/resource/lvm_volumegroupnodestatus.go
```diff
@@ -0,0 +1,70 @@
...
+	vgNodeStatusList := &lvmv1alpha1.LVMVolumeGroupNodeStatusList{}
+	if err := r.List(ctx, vgNodeStatusList, client.InNamespace(r.GetNamespace())); err != nil {
+		return fmt.Errorf("failed to list LVMVolumeGroupNodeStatus: %w", err)
+	}
+
+	for _, nodeItem := range vgNodeStatusList.Items {
+		if !nodeItem.GetDeletionTimestamp().IsZero() {
+			return fmt.Errorf("the LVMVolumeGroupNodeStatus %s is still present, waiting for deletion", nodeItem.GetName())
+		}
+		if err := r.Delete(ctx, &nodeItem); err != nil {
```
jakobmoellerdev, line 62: I think it would be wise to delete it only if it does not contain any volume groups anymore since that is managed by vgmanager. WDYT?

jakobmoellerdev (changes requested):
In general LGTM, but its missing unit and integration tests

internal/controllers/lvmcluster/resource/lvm_volumegroupnodestatus.go
suleymanakbas91: This comes after volume group removal (https://github.com/openshift/lvm-operator/pull/617/files#diff-3ddf36bfe9ab457a2b494fb56efbf5f1f70472c7e7f62d29779545f0a75ecff1R423) so the reconciler won't start removal of lvmvolumegroupnodestatus unless all the volume groups are gone. We could still add one more check here to double-check, but this will copy [this logic](https://github.com/openshift/lvm-operator/blob/main/internal/controllers/lvmcluster/resource/lvm_volumegroup.go#L93) so I don't know if it's worth it. Wdyt?

jakobmoellerdev: I believe the resource deletion should be indempotent when doing a concurrent delete. I know we currently delete serialized but when we eventually switch to concurrent removal of all resources then we would need the check.

jakobmoellerdev (approved):
/lgtm

jakobmoellerdev (approved):
/lgtm

PR #625, opened 2024-05-06
OCPEDGE-1085: Change the base image to fedora

suleymanakbas91:
Because of a recent bug (https://issues.redhat.com/browse/OCPBUGS-33244), we had to switch to a rhel base image which requires registry authentication. This is not ideal for development and is not accessible to external contributors. That's why we want to switch to a Fedora base image.


jakobmoellerdev (approved):
/lgtm

PR #628, opened 2024-05-16
OCPBUGS-32709: Add conditions to LVMCluster status

suleymanakbas91:
This PR adds two new conditions to LVMCluster CR: 
- `ResourcesAvailable`: Indicates whether the resources maintained by the operator are reconciled without any issues.
- `VolumeGroupsReady`: Indicates whether the volume groups maintained by the operator are in a ready state.

The state and the readiness of an LVMCluster CR is also now calculated using these conditions.


internal/controllers/lvmcluster/controller.go
```diff
@@ -210,6 +208,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (r *Reconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMCluster) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
+	instance.Status.Conditions = computeInitialConditions()
```
jakobmoellerdev, line 211: I suggest to change the condition type to https://github.com/openshift/lvm-operator/blob/ab9df4c80d2ac20eb9818d3ff44a337f2f8918ca/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1497 and then make use of these conditions

Then you can also use functions like https://github.com/openshift/lvm-operator/blob/0d154af54cbd6eef68fb5ee421ed4a37f9b8e017/vendor/k8s.io/apimachinery/pkg/api/meta/conditions.go#L31

WDYT?

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
+package lvmcluster
```
jakobmoellerdev, line 1: This entire file seems pretty huge for 2 simple conditions, why not use a template variable for the condition and then just deep copy this whenever you need to initialize them?

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
...
+	conditions = append(conditions, vgReadinessCondition)
+
+	return conditions
+}
+
+func computeResourcesSyncedCondition(init bool, reconciliationError error) lvmv1alpha1.Condition {
+	condition := lvmv1alpha1.Condition{
+		Type: lvmv1alpha1.ResourcesSynced,
+	}
+	if init {
```
jakobmoellerdev, line 48: This is not needed if you slightly restructure the code. You can initialize them by default and then modify the existing condition in later parts of the code. There is no need to compute them from scratch everytime.

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return condition
+}
+
+func computeVGReadinessCondition(init bool, vgNodeStatusList *lvmv1alpha1.LVMVolumeGroupNodeStatusList) lvmv1alpha1.Condition {
+	condition := lvmv1alpha1.Condition{
+		Type: lvmv1alpha1.VolumeGroupsReady,
+	}
+	if init {
```
jakobmoellerdev, line 72: why not just set this always and then override fields as necessary? makes init unnecessary

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
...
+	if isFailed {
+		condition.Status = corev1.ConditionFalse
+		condition.Reason = ReasonVGsFailed
+		condition.Message = MessageVGsFailed
+	} else if isDegraded {
+		condition.Status = corev1.ConditionFalse
+		condition.Reason = ReasonVGsDegraded
+		condition.Message = MessageVGsDegraded
+	} else if isProgressing {
+		condition.Status = corev1.ConditionFalse
```
jakobmoellerdev, line 103: you dont need to set default values everytime if you set them once above

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
...
+		Type: lvmv1alpha1.VolumeGroupsReady,
+	}
+	if init {
+		condition.Status = corev1.ConditionFalse
+		condition.Reason = ReasonVGReadinessInProgress
+		condition.Message = MessageVGReadinessInProgress
+		return condition
+	}
+
+	var isReady, isDegraded, isFailed, isProgressing bool
```
jakobmoellerdev, line 79: I think this is quite a complicated logic, I think we could just write a small function that translates VGStatus => Reason+Message in a function with a switch case

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+		)
+	}
+
+	return allVgStatuses
+}
+
+func computeReadiness(conditions []lvmv1alpha1.Condition) (state lvmv1alpha1.LVMStateType, ready bool) {
+	readyConditionCount := 0
+	for _, c := range conditions {
```
jakobmoellerdev, line 142: similarly to above but a bit more complex, You could just translate c.Reason to a State in a function and then this function would get significantly less complex

api/v1alpha1/lvmcluster_types.go
```diff
@@ -169,24 +169,11 @@ type Condition struct {
...
-	// It may still be available, but it is degraded
-	ConditionDegraded ConditionType = "Degraded"
-
-	// ConditionUpgradeable indicates whether the resources maintained by the operator are in a state that is safe to upgrade.
-	// When `False`, the resources maintained by the operator should not be upgraded and the
-	// message field should contain a human readable description of what the administrator should do to
-	// allow the operator to successfully update the resources maintained by the operator.
-	ConditionUpgradeable ConditionType = "Upgradeable"
+	// ResourcesSynced indicates whether the resources maintained by the operator are reconciled without any issues.
+	ResourcesSynced ConditionType = "ResourcesSynced"
```
jakobmoellerdev, line 173: Nit:
```suggestion
	ResourcesSynced ConditionType = "ResourcesAvailable"
```
We dont just check for reconcile without issues, we also check wether they are available. True, that IS part of the reconcile loop, but the current status "ResourcesSynced" is not really clear on what that means for the operator.

internal/controllers/lvmcluster/controller.go
```diff
@@ -194,7 +192,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	result, reconcileError := r.reconcile(ctx, lvmCluster)
 
-	statusError := r.updateLVMClusterStatus(ctx, lvmCluster)
+	statusError := r.updateLVMClusterStatus(ctx, lvmCluster, reconcileError)
```
jakobmoellerdev, line 195: Could you explain to me again Why its necessary to do an explicit Status error check here? I was expecting that part of reconcile would be to update the status. Now its still 2 different functions, and its hard to incorporate new status elements. Similarly the below check on the reconcileError is not synchronized with the statusError, we wanted to have this error be one I think?

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func computeLVMClusterReadiness(conditions []metav1.Condition) (lvmv1alpha1.LVMStateType, bool) {
+	state := lvmv1alpha1.LVMStatusProgressing
+	readyConditionCount := 0
+	for _, c := range conditions {
+		if c.Status == metav1.ConditionTrue {
+			readyConditionCount++
+		} else {
+			state = translateReasonToState(c.Reason, state)
```
jakobmoellerdev, line 154: why do we only do this when a condition is false? shouldnt we recompute the state everytime?

internal/controllers/lvmcluster/status.go
```diff
@@ -0,0 +1,174 @@
...
+func setVolumeGroupsReadyConditionDegraded(instance *lvmv1alpha1.LVMCluster) {
+	meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+		Type:    lvmv1alpha1.VolumeGroupsReady,
+		Status:  metav1.ConditionFalse,
+		Reason:  ReasonVGsDegraded,
+		Message: MessageVGsDegraded,
+	})
+}
+
+func computeVolumeGroupsReadyReason(vgNodeStatusList *lvmv1alpha1.LVMVolumeGroupNodeStatusList) string {
```
jakobmoellerdev, line 96:
```suggestion
func computeVolumeGroupsReadyReason(vgNodeStatusList *lvmv1alpha1.LVMVolumeGroupNodeStatusList) string {
	vgsCount := 0
	readyVGsCount := 0
	reason := ReasonVGReadinessInProgress

	for _, nodeItem := range vgNodeStatusList.Items {
		for _, vgStatus := range nodeItem.Spec.LVMVGStatus {
			vgsCount++
			switch vgStatus.Status {
			case lvmv1alpha1.VGStatusReady:
				readyVGsCount++
			case lvmv1alpha1.VGStatusFailed:
				return ReasonVGsFailed
			case lvmv1alpha1.VGStatusDegraded:
				reason = ReasonVGsDegraded
			}
		}
	}

	if readyVGsCount == vgsCount {
		return ReasonVGsReady
	}

	return reason
}
```

internal/controllers/lvmcluster/controller.go
```diff
@@ -253,128 +240,57 @@ func (r *Reconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMClu
...
-	if isFailed {
-		instance.Status.State = lvmv1alpha1.LVMStatusFailed
-	} else if isDegraded {
-		instance.Status.State = lvmv1alpha1.LVMStatusDegraded
-	} else if isReady && expectedVGCount <= readyVGCount { // readyVGCount can be greater than expectedVGCount in case there are taints on some nodes, e.g. during an upgrade. See https://issues.redhat.com/browse/OCPBUGS-30830 for further details.
-		instance.Status.State = lvmv1alpha1.LVMStatusReady
-		instance.Status.Ready = true
+	instance.Status.DeviceClassStatuses = computeDeviceClassStatuses(vgNodeStatusList)
+
+	reason := computeVolumeGroupsReadyReason(vgNodeStatusList)
```
jakobmoellerdev, line 276:
```suggestion
reason := computeVolumeGroupsReadyReason(vgNodeStatusList)

switch reason {
case ReasonVGsReady:
    setVolumeGroupsReadyConditionTrue(instance)
case ReasonVGsFailed:
    setVolumeGroupsReadyConditionFailed(instance)
case ReasonVGsDegraded:
    setVolumeGroupsReadyConditionDegraded(instance)
}
```

internal/controllers/lvmcluster/controller.go
```diff
@@ -253,128 +240,57 @@ func (r *Reconciler) reconcile(ctx context.Context, instance *lvmv1alpha1.LVMClu
...
 
 	return ctrl.Result{}, nil
 }
 
-func (r *Reconciler) updateLVMClusterStatus(ctx context.Context, instance *lvmv1alpha1.LVMCluster) error {
+func (r *Reconciler) updateLVMClusterStatus(ctx context.Context, instance *lvmv1alpha1.LVMCluster, reconcileError error) error {
 	logger := log.FromContext(ctx)
 
-	vgNodeMap := make(map[string][]lvmv1alpha1.NodeStatus)
+	if reconcileError != nil {
```
jakobmoellerdev, line 264: this if statement is not necessary if you simply set the resourcesAvailableCondition at line 241, then you can also drop passing the error and rename the function to setVolumeGroupsReadyCondition

jakobmoellerdev (commented):
only a few codesmells but general approach is LGTM :)

internal/controllers/lvmcluster/status.go
suleymanakbas91: I changed it to always recompute. I also added an Unknown state as default to distinguish an unknown state from in progress. Tell me what you think.

jakobmoellerdev (approved):
/lgtm

Great Job on that PR, that one is big for maintaining conditions :tada:

PR #631, opened 2024-06-10
OCPBUGS-33304: fix: Rework LVMVolumeGroupNodeStatus management

jakobmoellerdev:
When LVMVolumeGroupNodeStatus was originally created it was thought of to be fully managed by the LVMVolumeGroups reconciler in vg-manager. However this is problematic because there must always be 1 owner for full blockOwnerDeletion and Deletion Propagation to work in kubernetes (Controller Reference must be set from the Owner to the owning object). This is not possible with that approach. If we reimagine the LVMCluster to create LVMVolumeGroupNodeStatus instead (by precalculating the Nodes for which we should create a node status) we can achieve the following:
1. Create NodeStatus as soon as LVMCluster exists
2. Have a Finalizer that protects from dropping a NodeStatus prematurely (this can still lead to issues where vg-manager can no longer remove its own volume group from the status due to a set finalizer, but its better than LVMCluster assuming that the vg was removed correctly and showing a wrong status)
3. Have a hook that allows to delete NodeStatus elements when the LVMCluster is removed.

This will allow external projects such as console to do clean operand deletion as we will now have a way to build a clean dependency chain from LVMCluster down to the owned CRs.

### DIFF
--- a/internal/controllers/lvmcluster/resource/lvm_volumegroupnodestatus.go
+++ b/internal/controllers/lvmcluster/resource/lvm_volumegroupnodestatus.go
@@ -19,17 +19,20 @@ package resource
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	lvmv1alpha1 "github.com/openshift/lvm-operator/api/v1alpha1"
-
+	"github.com/openshift/lvm-operator/internal/controllers/lvmcluster/selector"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	cutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


internal/controllers/lvmcluster/resource/lvm_volumegroupnodestatus.go
```diff
@@ -38,67 +42,125 @@ func LVMVGNodeStatus() Manager {
...
+			}
+			return nil
+		})
+
+		if err != nil {
+			return fmt.Errorf("%s failed to reconcile: %w", l.GetName(), err)
+		}
+
+		if result != cutil.OperationResultNone {
+			logger.Info("LVMVolumeGroupNodeStatus applied to cluster", "operation", result, "name", lvmVGNodeStatus.Name)
```
suleymanakbas91, line 88: Should we add V(2) to the log here to make it consistent with the other resources?

jakobmoellerdev: Actually I think we have a bug here, I remember writing this some time back but I couldnt get the logs to run with V > 1. With (v1) they still show up in debug but not with V2. Here I forgot the V(2) back in but we probably should go through the code base and change every occurance to V(1) instead and verify the debug logs.

jakobmoellerdev: Changed to V1. need to open a follow up PR for the rest. Now the logs are visible with `--zap-log-level=debug` in the operator pod

PR #634, opened 2024-06-13
OCPEDGE-1151: chore: k8s 1.30 / go1.22

jakobmoellerdev:
Notable issues this rebase:
- Client go pinning: https://github.com/openshift/lvm-operator/pull/634/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R50 due to https://github.com/kubernetes-csi/external-provisioner/blob/master/go.mod#L23
- external-provisioner carry-patch: https://github.com/openshift/lvm-operator/pull/634/files#diff-8cfbce379f44d92fdddc5ea8b6c839474c863e3adcefc2dc14425ecc4482de17R7 due to external snapshotter libary being outdated
- metrics redefinition: https://github.com/openshift/lvm-operator/pull/634/files#diff-db7d93651239035ec6fa816e634d3fc7a2d6694b7c51e2e1f30b361e7a500765R60 in external provisioner forces us to register metrics different, being incompatible with controller runtime right now (PR for fixing that in upstream is up)
- external-provisioner go.mod: import only via commit hash possible, go.mod does not contain /v5 suffix (PR for fixing that in upstream is up)


cmd/vgmanager/vgmanager.go
```diff
@@ -150,7 +150,6 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 	if err := loadConfFile(ctx, lvmdConfig, lvmd.DefaultFileConfigPath); err != nil {
 		opts.SetupLog.Error(err, "lvmd config could not be loaded, starting without topolvm components and attempting bootstrap")
 	} else {
-		topoLVMD.Containerized(true)
 		lvclnt, vgclnt := topoLVMD.NewEmbeddedServiceClients(ctx, lvmdConfig.DeviceClasses, lvmdConfig.LvcreateOptionClasses)
```
suleymanakbas91, line 153: Why do we remove this?

jakobmoellerdev: topolvm has removed this flag because the nsenter call is a noop if not run in a container so they dont need the check anymore

PR #636, opened 2024-06-19
OCPBUGS-35817: LVMCluster shows Failed if it could not create any VGs

suleymanakbas91:


internal/controllers/lvmcluster/status.go
```diff
@@ -69,6 +69,10 @@ func setResourcesAvailableConditionFalse(instance *lvmv1alpha1.LVMCluster, recon
 }
 
 func setVolumeGroupsReadyCondition(instance *lvmv1alpha1.LVMCluster, vgNodeStatusList *lvmv1alpha1.LVMVolumeGroupNodeStatusList) {
+	if len(vgNodeStatusList.Items) < 1 {
```
jakobmoellerdev, line 72: I think this check is too naive. we should check that for every vg in the device class we have all the volume groups in the node status that we would expect based on the node selector. For example now, if no deviceClasses are present, this would be a failed status, but with the Microshift integration this is a valid use case.

internal/controllers/lvmcluster/status.go
```diff
@@ -210,3 +183,113 @@ func translateReasonToState(reason string, currentState lvmv1alpha1.LVMStateType
...
+				deviceClass.Name)
+		}
+	}
+
+	return nil
+}
+
+func isNodeValid(node *corev1.Node, cluster *lvmv1alpha1.LVMCluster, deviceClass *lvmv1alpha1.DeviceClass) (bool, error) {
+	// Check if node tolerates all taints
+	if !selector.ToleratesAllTaints(node.Spec.Taints, cluster.Spec.Tolerations) {
```
jakobmoellerdev, line 280: what happens if taints or tolrations are nil?

jakobmoellerdev (commented):
have to go for a second round on the test cases but mostly LGTM

internal/controllers/lvmcluster/status.go
suleymanakbas91: I actually have a test case for this: https://github.com/openshift/lvm-operator/pull/636/files#diff-c50dd555085f94f16659d15a7a8e47f8b59c9ce6531ca5a858bd1c7309662df5R51

jakobmoellerdev (approved):
/lgtm

PR #642, opened 2024-06-25
OCPEDGE-1165: fix: Ensure no racy CSI plugin registration, better startup behavior for vgmanager, failing test summary artifact

jakobmoellerdev:
This stops vgmanager from going healthy too early as we need to signal healthiness only when the driver has been started. Fixes the racy registration server that now properly starts and shuts down and is part of the healthiness check we use in vgmanager. That will ensure that no vgmanager gets ready that isnt registered in the kubelet. Ensures that vgmanager pods are gone before LVMCluster is freed.

Adds an artifact into the e2e step on failure that can be used to see the immediate LVMS cluster state when a failure occurs in case of a cleanup in the post failure steps.

Validates in tests that all CSINodes have proper registrations for vgmanager so we dont have weird flakes in the future.

### DIFF
--- a/cmd/vgmanager/vgmanager.go
+++ b/cmd/vgmanager/vgmanager.go
@@ -182,7 +182,7 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 		registrationServer := internalCSI.NewRegistrationServer(
 			constants.TopolvmCSIDriverName, registrationPath(), []string{"1.0.0"})
 		registerapi.RegisterRegistrationServer(grpcServer, registrationServer)
-		if err = mgr.Add(runners.NewGRPCRunner(grpcServer, pluginRegistrationSocketPath(), false)); err != nil {
+		if err = mgr.Add(internalCSI.NewGRPCRunner(grpcServer, pluginRegistrationSocketPath(), false)); err != nil {
 			return fmt.Errorf("could not add grpc runner for registration server: %w", err)
 		}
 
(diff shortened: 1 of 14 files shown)


jakobmoellerdev:
/hold
putting on hold until Ive confirmed the flakiness is gone and I added repeat tests that we dont want to have on main

cmd/vgmanager/vgmanager.go
```diff
@@ -216,7 +219,18 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 		return fmt.Errorf("unable to set up ready check: %w", err)
 	}
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+	if err := mgr.AddHealthzCheck("healthz", func(req *http.Request) error {
+		if len(lvmdConfig.DeviceClasses) == 0 {
+			log.FromContext(req.Context()).Error(ErrNoDeviceClassesAvailable, "not healthy",
+				"available_device_classes", lvmdConfig.DeviceClasses)
```
suleymanakbas91, line 225: We already know that len(lvmdConfig.DeviceClasses) = 0, so why do we want to print it here?

cmd/vgmanager/vgmanager.go
```diff
@@ -335,3 +352,10 @@ func readyCheck(mgr manager.Manager) healthz.Checker {
 		return nil
 	}
 }
+
+func newGRPCServer() *grpc.Server {
+	return grpc.NewServer(grpc.UnaryInterceptor(ErrorLoggingInterceptor),
+		grpc.SharedWriteBuffer(true),
+		grpc.MaxConcurrentStreams(2),
```
suleymanakbas91, line 359: Could you please add a comment on why we're setting MaxConcurrentStreams and NumStreamWorkers to 2?

suleymanakbas91 (commented):
LGTM, just small nits

PR #683, opened 2024-08-05
OCPBUGS-37711: fix: use marker annotation to determine successful wipe and rewipe on failure

jakobmoellerdev:


internal/controllers/vgmanager/wipe_devices.go
```diff
@@ -4,62 +4,96 @@ import (
...
+		} else if deviceWiped {
+			updated = true
 		}
-		if deviceWiped {
-			wiped = true
+	}
+
+	if updated {
+		if volumeGroup.Annotations[constants.DevicesWipedAnnotation] == "" {
+			volumeGroup.Annotations = make(map[string]string)
```
suleymanakbas91, line 47: Are we overriding all the annotations here?

suleymanakbas91 (changes requested):

internal/controllers/vgmanager/wipe_devices.go
jakobmoellerdev: I had a brain lag and didnt compare for map==nil but for the key, that should be fixed now

internal/controllers/vgmanager/wipe_devices.go
```diff
@@ -4,62 +4,96 @@ import (
...
 			logger.Info(fmt.Sprintf("skipping wiping optional device %s: %v", path, err))
+		} else if deviceWiped {
+			updated = true
 		}
-		if deviceWiped {
-			wiped = true
+	}
+
+	if updated {
+		if volumeGroup.Annotations[constants.DevicesWipedAnnotation] == "" {
```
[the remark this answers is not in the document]
jakobmoellerdev: see above, thanks for calling out

PR #690, opened 2024-08-09
OCPEDGE-1168: change DevicePath field to its own type 

qJkee:
Introduce a new type called DevicePath to contain user provided device path. This is done in order to create a standard way of working with resolved if symlink or raw(unresolved) device path and avoid errors when it's not resolved while it's expected to be.

### DIFF
--- a/api/v1alpha1/lvmcluster_types.go
+++ b/api/v1alpha1/lvmcluster_types.go
@@ -130,17 +130,23 @@ type DeviceSelector struct {
 
 	// Paths specify the device paths.
 	// +optional
-	Paths []string `json:"paths,omitempty"`
+	Paths []DevicePath `json:"paths,omitempty"`
 
 	// OptionalPaths specify the optional device paths.
 	// +optional
-	OptionalPaths []string `json:"optionalPaths,omitempty"`
+	OptionalPaths []DevicePath `json:"optionalPaths,omitempty"`
 
 	// ForceWipeDevicesAndDestroyAllData is a flag to force wipe the selected devices.
 	// This wipes the file signatures on the devices. Use this feature with caution.
 	// Force wipe the devices only when you know that they do not contain any important data.
 	// +optional
 	ForceWipeDevicesAndDestroyAllData *bool `json:"forceWipeDevicesAndDestroyAllData,omitempty"`
+}
+
+type DevicePath string
+
+func (d DevicePath) Unresolved() string {
+	return string(d)
 }
 
 type LVMStateType string
--- a/cmd/vgmanager/vgmanager.go
+++ b/cmd/vgmanager/vgmanager.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 	"time"
 
@@ -192,17 +193,18 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 	}
 
 	if err = (&vgmanager.Reconciler{
-		Client:        mgr.GetClient(),
-		EventRecorder: mgr.GetEventRecorderFor(vgmanager.ControllerName),
-		LVMD:          lvmd.DefaultConfigurator(),
-		Scheme:        mgr.GetScheme(),
-		LSBLK:         lsblk.NewDefaultHostLSBLK(),
-		Wipefs:        wipefs.NewDefaultHostWipefs(),
-		Dmsetup:       dmsetup.NewDefaultHostDmsetup(),
-		LVM:           lvm.NewDefaultHostLVM(),
-		NodeName:      nodeName,
-		Namespace:     operatorNamespace,
-		Filters:       filter.DefaultFilters,
+		Client:           mgr.GetClient(),
+		EventRecorder:    mgr.GetEventRecorderFor(vgmanager.ControllerName),
+		LVMD:             lvmd.DefaultConfigurator(),
+		Scheme:           mgr.GetScheme(),
+		LSBLK:            lsblk.NewDefaultHostLSBLK(),
+		Wipefs:           wipefs.NewDefaultHostWipefs(),
+		Dmsetup:          dmsetup.NewDefaultHostDmsetup(),
+		LVM:              lvm.NewDefaultHostLVM(),
+		NodeName:         nodeName,
+		Namespace:        operatorNamespace,
+		Filters:          filter.DefaultFilters,
+		SymlinkResolveFn: filepath.EvalSymlinks,
 	}).SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("unable to create controller VGManager: %w", err)
 	}
--- a/docs/design/lvm-operator-manager.md
+++ b/docs/design/lvm-operator-manager.md
@@ -61,7 +61,7 @@ The Operator requires elevated permissions to interact with the host's LVM comma
 
 ## Implementation Notes
 
-Each unit of reconciliation should implement the `reconcileUnit` interface. This is run by the controller. Errors and success messages are propagated as Operator status and events. This interface is defined in [lvmcluster_controller.go](../../internal/controllers/lvmcluster/lvmcluster_controller.go)
+Each unit of reconciliation should implement the `reconcileUnit` interface. This is run by the controller. Errors and success messages are propagated as Operator status and events. This interface is defined in [manager.go](../../internal/controllers/lvmcluster/resource/manager.go)
 
 ```go
 type resourceManager interface {
--- a/internal/controllers/symlink-resolver/resolver.go
+++ b/internal/controllers/symlink-resolver/resolver.go
@@ -1,0 +1,42 @@
+package symlinkResolver
+
+import (
+	"path/filepath"
+	"sync"
+)
+
+type Resolver struct {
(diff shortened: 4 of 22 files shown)


api/v1alpha1/lvmcluster_types.go
```diff
@@ -143,6 +143,18 @@ type DeviceSelector struct {
 	ForceWipeDevicesAndDestroyAllData *bool `json:"forceWipeDevicesAndDestroyAllData,omitempty"`
 }
 
+type resolverFunc func(string) (string, error)
+
+type DevicePath string
```
jakobmoellerdev, line 148: What happens if Resolve gets called more than once? Is there a reason this is not a sync.OnceFunc? AFAIU, this is gonna be the same behavior as the current code, but I thought we agreed that it makes sense to only resolve once per reconcile

test/performance/parse.go
```diff
@@ -31,7 +31,7 @@ func parseMetrics(metricsResult string) ([]RawMetric, error) {
 		return nil, fmt.Errorf("failed to unmarshal metrics: %w", err)
 	}
 
-	metrics := []RawMetric{}
+	metrics := make([]RawMetric, 0)
```
jakobmoellerdev, line 34:
```suggestion
	var metrics []RawMetric
```

jakobmoellerdev (commented):
Just a tad confused on the actual impact on the change, as I think we said that we want to make sure that resolution happens on a At Most Once basis. Other than that LGTM

docs/design/lvm-operator-manager.md
```diff
@@ -61,7 +61,7 @@ The Operator requires elevated permissions to interact with the host's LVM comma
 
 ## Implementation Notes
 
-Each unit of reconciliation should implement the `reconcileUnit` interface. This is run by the controller. Errors and success messages are propagated as Operator status and events. This interface is defined in [lvmcluster_controller.go](../../internal/controllers/lvmcluster/lvmcluster_controller.go)
+Each unit of reconciliation should implement the `reconcileUnit` interface. This is run by the controller. Errors and success messages are propagated as Operator status and events. This interface is defined in [lvmcluster_controller.go](../../internal/controllers/lvmcluster/resource/manager.go)
```
suleymanakbas91, line 64:
```suggestion
Each unit of reconciliation should implement the `reconcileUnit` interface. This is run by the controller. Errors and success messages are propagated as Operator status and events. This interface is defined in [manager.go](../../internal/controllers/lvmcluster/resource/manager.go)
```

internal/controllers/vgmanager/devices.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/lvm-operator/v4/api/v1alpha1"
```
suleymanakbas91, line 23: Could you move this import to the github group in a few lines below?

internal/controllers/vgmanager/kname_test.go
```diff
@@ -1,17 +1,18 @@
 package vgmanager
 
 import (
+	"github.com/openshift/lvm-operator/v4/api/v1alpha1"
```
suleymanakbas91, line 4: Same here. 

We try to keep a consistent import order throughout the codebase:
- golang imports
- github imports
- k8s imports
(with new lines in between)

suleymanakbas91 (changes requested):

cmd/vgmanager/vgmanager.go
```diff
@@ -192,17 +193,18 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
...
+		LVMD:            lvmd.DefaultConfigurator(),
+		Scheme:          mgr.GetScheme(),
+		LSBLK:           lsblk.NewDefaultHostLSBLK(),
+		Wipefs:          wipefs.NewDefaultHostWipefs(),
+		Dmsetup:         dmsetup.NewDefaultHostDmsetup(),
+		LVM:             lvm.NewDefaultHostLVM(),
+		NodeName:        nodeName,
+		Namespace:       operatorNamespace,
+		Filters:         filter.DefaultFilters,
+		SymlinkResolver: symlink_resolver.NewWithDefaultResolver(),
```
jakobmoellerdev, line 207: Why do we need this across all reconciles?

internal/controllers/symlink-resolver/resolver.go
```diff
@@ -0,0 +1,43 @@
+package symlinkResolver
```
jakobmoellerdev, line 1: I suggest that all of the imports also use this case instead of _ . Waiting for input from @suleymanakbas91

internal/controllers/symlink-resolver/resolver.go
```diff
@@ -0,0 +1,43 @@
...
+		val, err = r.resolveFn(path)
+		if err != nil {
+			return "", err
+		}
+		r.cache.Store(path, val)
+	}
+	return val.(string), nil
+}
+
+func (r *Resolver) Cleanup() {
```
jakobmoellerdev, line 41: I dont think a resolver should be reusable and we should initiate it once per reconcile. If we think of resolver as Closeable, then this would look like:

```suggestion
func (r *Resolver) Close() error {
   r.cache = nil
   return nil
}
```

For GC thats a no-op so we might just completely drop this and initialize it from scratch in every reconcile from the resolveFn


Im also not sure if we need a sync.Map{} here, we are not in performance critical code, but this is more of a nit, so feel free to ignore.

internal/controllers/vgmanager/controller.go
```diff
@@ -105,6 +108,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	logger := log.FromContext(ctx)
 	logger.V(1).Info("reconciling")
 
+	// remove old entries because symlink can change
+	defer r.SymlinkResolver.Cleanup()
```
jakobmoellerdev, line 112: I think we could just initialize a new map and let gc do the rest. then you dont need a cleanup or a close routine actually (you can completely drop that code and compiled behavior is the same in gc because you assign to zero value in Cleanup())

api/v1alpha1/lvmcluster_types.go
```diff
@@ -143,6 +143,12 @@ type DeviceSelector struct {
 	ForceWipeDevicesAndDestroyAllData *bool `json:"forceWipeDevicesAndDestroyAllData,omitempty"`
 }
 
+type DevicePath string
```
jakobmoellerdev, line 146: Why do we need this type now? This only has one method which returns d the path as a string

jakobmoellerdev (commented):
Mostly nits, approach is LGTM for me.

Looking at the code for the resolver, it feels a little like we do not need the concurrent sync.Map access since we run in the reconcile and the reconcile doesnt have a concurrent call for device resolution AFAIK. (please correct me if im wrong)

Im also not sure if we should keep the DevicePath type because now its literally a string wrapper, although I must admit that the Unresolved() method makes it clear what is to be expected from that value. maybe a godoc would resolve that?

cmd/vgmanager/vgmanager.go
jakobmoellerdev: You can still do the same if you change the injection to the symlink resolution function alone, not the entire resolver object, e.g https://github.com/openshift/lvm-operator/pull/690/files#diff-08a2a7b11748abc371e0feb4f8983dc47562ed33dfbc1861940046a5d14d984aR15 can be injected as a function

```suggestion
		SymlinkFn: filepath.EvalSymlinks,
```

then you can exchange it in the tests

api/v1alpha1/lvmcluster_types.go
jakobmoellerdev: Pretty unidiomatic, but I agree with you. This should protect from unintended usage.

internal/controllers/symlink-resolver/resolver.go
jakobmoellerdev: sounds okay to me even though it feels like premature optimization. resolving

internal/controllers/vgmanager/controller_test.go
```diff
@@ -844,11 +858,10 @@ func testReconcileFailure(ctx context.Context) {
...
 
 	By("triggering wipefs failure", func() {
-		evalSymlinks = func(path string) (string, error) {
-			return path, nil
-		}
+		instances.Reconciler.SymlinkResolveFn = func(path string) (string, error) { return path, nil }
+
 		defer func() {
-			evalSymlinks = filepath.EvalSymlinks
+			instances.Reconciler.SymlinkResolveFn = filepath.EvalSymlinks
```
jakobmoellerdev, line 864: now that we no longer set a global variable, can we set it in the instances.Reconciler creation? that way we dont have to defer reset and we dont need to set them up everytime. WDYT?

jakobmoellerdev: All tests should work with that symlink resolution so you can do it once per suite. There is no need to revert it at all I believe

test/performance/parse.go
jakobmoellerdev: I think this is the last nit pending

jakobmoellerdev (commented):
/lgtm

PR #700, opened 2024-08-27
NO-JIRA: fix: kubelet csi registration watchdog

jakobmoellerdev:
adds a registration watchdog that kills vgmanager if a timeout has passed in case the CSI node registration leads nowhere.

Reasoning is that in CI we regularly create CSI Nodes extremely quickly and sometimes kubelets desired state of world populator goes into deadlock because of this. This restart should give kubelet a nudge by recreating the registration socket file on a new startup, causing the tests to continue normally.


jakobmoellerdev:
/hold

seems even an automatic restart is not enough for the kubelet to kick off the registration.

PR #701, opened 2024-08-28
OCPBUGS-13236: fix: allow deploy on cluster without prometheus

jakobmoellerdev:
This PR runs prometheus kustomization applies separately from the base kustomization with the makefile ignore `-` instruction, causing make to ignore any errors from applying to a cluster that does not have the monitoring CRDs installed (e.g. non-openshift).
Also makes the debug deploy easier to maintain by removing duplicated parts in the makefile


jakobmoellerdev:
We can but this will make the installation round trip longer. Do you prefer it? It is trivial to change I believe

PR #707, opened 2024-09-04
[DO NOT MERGE] fix: set termination grace period to 0

jakobmoellerdev:
Attempts to set termination grace to 0 to see if killing the pod helps with the registration races


jakobmoellerdev:
/test e2e-aws-single-node
the duplicate clusters test is unrelated of that flake and that is easy to fix, for now that can be ignored

PR #708, opened 2024-09-05
OCPEDGE-1247: make thin pool config editable

qJkee:
Make editable overprovisionRatio field in thin pool device class at lvm cluster


internal/controllers/vgmanager/lvmd/lvmd.go
```diff
@@ -31,6 +32,50 @@ const (
 	maxReadLength            = 2 * 1 << 20 // 2MB
 )
 
+func DeepCopyConfig(c *Config) *Config {
```
jakobmoellerdev, line 35: Why do you need this? Couldnt you just deref and reref the Config? Or use deepcopy gen?

jakobmoellerdev (commented):
Just a comment regarding the custom DeepCopy other than that looks okay to me.

A general question: How does TopoLVM behave when overprovisioning is reduced to a limit that is already reached?

jakobmoellerdev:
What do you mean with it works? It should not be able to overprovision. If that is not prohibited then we need to fix that in upstream

PR #727, opened 2024-09-23
OCPEDGE-1273: chore: k8s 1.31 dependency bumps 

jakobmoellerdev:
Noteworthy changes:

- Role definitions in generated RBAC are now aggregated
- Some missing mandatory fields in the CRDs were patched in that were previously accidentally removed
- Documentation for the update process was added
- One of the mandatory replace statements was able to be removed
- The release includes the latest state of topolvm which has a filesystem expansion issue when using an older version of topolvm (see https://github.com/topolvm/topolvm/releases/tag/v0.33.0 for details)


cmd/vgmanager/vgmanager.go
```diff
@@ -197,9 +196,6 @@ func run(cmd *cobra.Command, _ []string, opts *Options) error {
 			return fmt.Errorf("could not add topolvm metrics: %w", err)
 		}
 
-		if err := os.MkdirAll(topolvm.DeviceDirectory, 0755); err != nil {
```
suleymanakbas91, line 200: Why don't we need this anymore?

jakobmoellerdev: Because of https://github.com/topolvm/topolvm/releases/tag/v0.33.0 getting rid of the topolvm device directory

jakobmoellerdev: Concretely, this is the change: https://github.com/topolvm/topolvm/pull/930/files

PR #728, opened 2024-09-24
OCPEDGE-1230: add metadata size field

qJkee:


api/v1alpha1/lvmcluster_types.go
```diff
@@ -87,6 +91,10 @@ var (
 	ChunkSizeMaximum = resource.MustParse("1Gi")
 )
 
+var (
+	ThinPoolMetadataSizeDefault = resource.MustParse("1Gi")
```
jakobmoellerdev, line 95: This can be defaulted within the API type I believe

api/v1alpha1/lvmcluster_types.go
```diff
@@ -68,6 +68,10 @@ type ThinPoolConfig struct {
 	// It can be between 64Ki and 1Gi due to the underlying limitations of lvm2.
 	// +optional
 	ChunkSize *resource.Quantity `json:"chunkSize,omitempty"`
+
+	// MetadataSize specifies metadata size for thin pool.
+	// +optional
+	MetadataSize *resource.Quantity `json:"metadataSize,omitempty"`
```
jakobmoellerdev, line 74: This is missing an explanation on what is used when it is nil as well as minimum and maximum sizes

config/manager/kustomization.yaml
```diff
@@ -7,7 +7,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/lvms_dev/lvms-operator
+  newName: quay.io/bzamalut/lvms
```
jakobmoellerdev, line 10: Needs to be removed before merge

internal/controllers/vgmanager/lvm/lvm.go
```diff
@@ -489,6 +489,7 @@ func (hlvm *HostLVM) CreateLV(ctx context.Context, lvName, vgName string, sizePe
 		args = append(args, "-c", fmt.Sprintf("%vb", chunkSizeBytes))
 	}
 
+	args = append(args, "--poolmetadatasize", fmt.Sprintf("%vb", metadataSizeBytes))
```
jakobmoellerdev, line 492: this is missing a case if the metadataSizeBytes are not provided

internal/controllers/vgmanager/controller.go
```diff
@@ -616,7 +616,13 @@ func (r *Reconciler) addThinPoolToVG(ctx context.Context, vgName string, config
 	}
 
 	logger.Info("creating lvm thinpool")
-	if err := r.LVM.CreateLV(ctx, config.Name, vgName, config.SizePercent, convertChunkSize(config)); err != nil {
+
+	metadataSize := lvmv1alpha1.ThinPoolMetadataSizeDefault.Value()
```
jakobmoellerdev, line 620: Why do you default to 1Gi here instead of using the Hosts default value from lvm.conf?

internal/controllers/vgmanager/controller.go
```diff
@@ -689,6 +704,26 @@ func (r *Reconciler) extendThinPool(ctx context.Context, vgName string, lvSize s
...
+
+	currentMetadataSize, err := strconv.ParseInt(lvMetadataSize[:len(lvMetadataSize)-1], 10, 64)
+	if err != nil {
+		return fmt.Errorf("failed to parse metadata size. %w", err)
+	}
+	if currentMetadataSize >= newMetadataSize {
+		return nil
+	}
+
+	err = r.LVM.ExtendThinPoolMetadata(ctx, lvName, vgName, newMetadataSize)
```
suleymanakbas91, line 720: I believe this method should only do the validation as it is part of `validateLVs`. The actual extend operation can take place within `extendThinPool` called [here](https://github.com/openshift/lvm-operator/blob/3ca641782ca8ccfd7637d8c385ace913884459e7/internal/controllers/vgmanager/controller.go#L611) or just after it as they are closely related.

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -56,6 +56,7 @@ var (
 	ErrEmptyPathsWithMultipleDeviceClasses                   = errors.New("path list should not be empty when there are multiple deviceClasses")
 	ErrDuplicateLVMCluster                                   = errors.New("duplicate LVMClusters are not allowed, remove the old LVMCluster or work with the existing instance")
 	ErrThinPoolConfigCannotBeChanged                         = errors.New("ThinPoolConfig can not be changed")
+	ErrThinPoolMetadataSizeCanBeOnlyIncreased                = errors.New("thin pool metadata size can be only increased")
```
suleymanakbas91, line 59:
```suggestion
	ErrThinPoolMetadataSizeCanOnlyBeIncreased                = errors.New("thin pool metadata size can only be increased")
```

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -197,6 +203,21 @@ func (v *lvmClusterValidator) ValidateUpdate(_ context.Context, old, new runtime
 			} else if !reflect.DeepEqual(newThinPoolConfig.ChunkSize, oldThinPoolConfig.ChunkSize) {
 				return warnings, fmt.Errorf("ThinPoolConfig.ChunkSize is invalid: %w", ErrThinPoolConfigCannotBeChanged)
 			}
+
+			if newThinPoolConfig.MetadataSizeCalculationPolicy == MetadataSizePolicyStatic {
```
suleymanakbas91, line 207: Do we allow switching between Host and Static policies during an update?

suleymanakbas91 (changes requested):

PR #791, opened 2024-11-19
OCPEDGE-1593: feat(controller): skip controller name validation

qJkee:
With new controller runtime kubernetes introduced controller name check. Because of that, graceful restarts started to fail. Because there are no other workaround - we simply disable this check


suleymanakbas91:
@qJkee is this still needed?

suleymanakbas91:
/override "Red Hat Konflux / lvm-operator-enterprise-contract / lvm-operator"
/override "Red Hat Konflux / lvm-operator-main-integration-tests / lvm-operator"

PR #826, opened 2024-12-20
[release-4.12] OCPBUGS-46622: Bump golang.org/x/net to v0.33.0

suleymanakbas91:


eggfoobar:
4.13 is not being released, so the dependent bug is in `Obsolete` status, removing invalid bug label.

/remove-label jira/invalid-bug

PR #884, opened 2025-02-27
OCPEDGE-1588: feat: enterprise contract warnings updates

jeff-roche:
Adding in missing tasks that will be required soon as well as updated task image references so we don't need to continuously update after bug fixes

### DIFF
--- a/.tekton/multi-arch-build-pipeline.yaml
+++ b/.tekton/multi-arch-build-pipeline.yaml
@@ -17,7 +17,7 @@ spec:
       - name: name
         value: init
       - name: bundle
-        value: quay.io/konflux-ci/tekton-catalog/task-init:0.2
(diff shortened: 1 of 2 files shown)


suleymanakbas91:
I see that we now have lots of EC failures with `Task apply-tags is used by pipeline task apply-tags via an unpinned reference.`.

jeff-roche:
> I see that we now have lots of EC failures with `Task apply-tags is used by pipeline task apply-tags via an unpinned reference.`.

@suleymanakbas91 I guess we will need to continuously monitor the EC jobs and update the SHAs :big-sad:

jeff-roche:
It doesn't appear to use the new pipeline config from the PR, it must only use the one that is committed to main because it has all the old image refs and still complains the new steps are missing.

jeff-roche:
@suleymanakbas91 this is ready for another look, ignore the EC results due to ^

PR #924, opened 2025-03-24
fix: bundle tagging

jeff-roche:
- Add the apply-tags task to the bundle pipeline
- Add the build-image-index task the bundle pipeline


jeff-roche:
/override ci/prow/ci-index-lvm-operator-bundle ci/prow/images ci/prow/precommit-check

PR #955, opened 2025-04-29
OCPEDGE-1784: change default namespace

qJkee:


bundle/manifests/lvms-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -13,14 +13,14 @@ spec:
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: lvms-operator-metrics-service.openshift-storage.svc
+      serverName: lvms-operator-metrics-service.openshift-lvm-storage.svc
```
suleymanakbas91, line 16: I wonder if this causes a case where the users may not see the metrics after an upgrade.

suleymanakbas91 (changes requested):

PR #1038, opened 2025-06-24
OCPBUGS-58059: feat(controller): check for stale finalizers

qJkee:
Cleanup lvm related objects finalizers in case
if node is gone or become unhealthy


internal/controllers/lvmcluster/controller.go
```diff
@@ -329,18 +334,31 @@ func (r *Reconciler) setRunningPodImage(ctx context.Context) error {
 	return nil
 }
 
-func (r *Reconciler) logicalVolumesExist(ctx context.Context) (bool, error) {
+// TODO: remove logical volume finalizer if node is no longer healthy
```
jeff-roche, line 337: Should we remove this todo comment?

PR #1048, opened 2025-06-25
NO-JIRA: chore: updating renovate config

jeff-roche:


jeff-roche:
/override ci/prow/precommit-check ci/prow/images ci/prow/ci-index-lvm-operator-bundle

PR #1123, opened 2025-07-16
OCPBUGS-51139: upgrade openshift/topolvm

qJkee:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/lvm-operator/v4
 
-go 1.24.3
```
jeff-roche, line 3: We should be on 1.24 for this development stream

PR #1147, opened 2025-07-23
OCPEDGE-1980: feat(config): add network policies

qJkee:
Add network policies to control networking for lvms pods


bundle/manifests/lvms-operator-network-policy_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,41 @@
...
+spec:
+  egress:
+  - ports:
+    - port: 6443
+      protocol: TCP
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: system
```
suleymanakbas91, line 18: When we deploy this through OLM, will it remain as `system`? It looks similar to what we had by using the hard-coded namespace in the service monitor.

suleymanakbas91 (changes requested):

PR #1150, opened 2025-07-28
OCPBUGS-56352: feat(resource): deploy ServiceMonitor as resource

qJkee:
deploy ServiceMonitor from operator instead of manifest to support custom namespace installation via OLM


internal/controllers/lvmcluster/resource/servicemonitor.go
```diff
@@ -0,0 +1,175 @@
...
+func (s serviceMonitorManager) IsPrometheusAvailable(r Reconciler, ctx context.Context) (bool, error) {
+	logger := log.FromContext(ctx).WithValues("resourceManager", s.GetName())
+
+	logger.Info("Checking if PrometheusCRD is available")
+	// check if the ServiceMonitor CRD is available
+	if err := r.Get(ctx, types.NamespacedName{Name: "servicemonitors.monitoring.coreos.com"}, &apiextensionsv1.CustomResourceDefinition{}); err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			return false, fmt.Errorf("failed to get ServiceMonitor CRD: %w", err)
+		}
+		logger.V(2).Info("ServiceMonitor CRD not available, skipping ServiceMonitor creation")
```
suleymanakbas91, line 170: This log is redundant as we have the same log where we call the function.

suleymanakbas91 (changes requested):

PR #1163, opened 2025-08-01
[release-4.19]OCPBUGS-59177: feat(resource): deploy ServiceMonitor as resource

qJkee:
deploy ServiceMonitor from operator instead of manifest to support custom namespace installation via OLM


jeff-roche:
Accidentally merged the wrong PR

PR #1164, opened 2025-08-04
NO-JIRA: fix: adding the images-mirror-set for tekton tasks

jeff-roche:


jeff-roche:
/override ci/prow/images ci/prow/ci-index-lvm-operator-bundle ci/prow/precommit-check

PR #1488, opened 2025-10-07
NO-JIRA: feat: adding a catalog build pipeline so we can manage task references

jeff-roche:


jeff-roche:

PR #1630, opened 2025-11-03
NO-JIRA: add setup-lvms-prereq slash command

kasturinarra:
This adds a Claude Code slash command to help users set up prerequisites
for testing unreleased LVMS operator builds on OpenShift clusters. It
 supports both connected and disconnected cluster deployments.


jeff-roche:
Out of curiosity, how does this differ from the scripts I've created on the release-management branch? (other than using claude code). As far as I can tell, this does the same thing

PR #1657, opened 2025-11-07
OCPEDGE-2194: feat(device class): allow device class removal on day2

qJkee:
Allow removing device class from LVMCluster on day 2

### DIFF
--- a/api/v1alpha1/lvmcluster_webhook.go
+++ b/api/v1alpha1/lvmcluster_webhook.go
@@ -174,10 +174,10 @@ func (v *lvmClusterValidator) ValidateUpdate(_ context.Context, old, new runtime
 		return warnings, fmt.Errorf("failed to parse LVMCluster")
 	}
 
-	// Validate all the old device classes still exist
-	err = validateDeviceClassesStillExist(oldLVMCluster.Spec.Storage.DeviceClasses, l.Spec.Storage.DeviceClasses)
+	// Validate device class removal follows the business rules
+	err = validateDeviceClassRemoval(oldLVMCluster.Spec.Storage.DeviceClasses, l.Spec.Storage.DeviceClasses)
 	if err != nil {
-		return warnings, fmt.Errorf("invalid: device classes were deleted from the LVMCluster: %w", err)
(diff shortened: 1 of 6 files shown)


api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -265,20 +265,25 @@ func (v *lvmClusterValidator) ValidateUpdate(_ context.Context, old, new runtime
...
+		newDeviceClassMap[deviceClass.Name] = deviceClass
 	}
 
-	// if any old device class is removed now
-	if len(deviceClassMap) != 0 {
-		return fmt.Errorf("device classes can not be removed from the LVMCluster once added oldDeviceClasses:%v, newDeviceClasses:%v", old, new)
+	for _, oldDeviceClass := range old {
+		if _, exists := newDeviceClassMap[oldDeviceClass.Name]; !exists {
+			if oldDeviceClass.Default {
+				return fmt.Errorf("cannot delete default device class")
```
suleymanakbas91: It'd be helpful to include the deleted default device class name in the error message.

suleymanakbas91 (changes requested):

internal/controllers/lvmcluster/controller.go
```diff
@@ -337,6 +345,84 @@ func (r *Reconciler) setRunningPodImage(ctx context.Context) error {
...
+
+		if _, ok := desiredVgNames[currentVG.Name]; ok {
+			continue
+		}
+
+		logger.Info("deleting lvm volume group and storage class because it no longer exists in device class list", "name", currentVG.Name)
+
+		err := r.Delete(ctx, &currentVG)
+		if err != nil {
+			return fmt.Errorf("failed to delete orphaned LVMVolumeGroup %s: %w", currentVG.Name, err)
```
pacevedom: Returning early when unable to delete orphans might prevent storageclasses or volume snapshot classes without cleaning up, should we keep going and add it to an error list?

qJkee: That's a good point. I think we need to have OwnerReferences on objects so we can delete only one main object rather than deleting them separately

PR #1726, opened 2025-11-21
NO-JIRA: fix: updated rpm lockfiles and changed the must-gather base image

jeff-roche:


jeff-roche:
cc @pacevedom @qJkee @mmakwana30

PR #1870, opened 2025-12-19
OCPEDGE-2015: Add SNO QE tests

mmakwana30:


test/integration/qe-tests/lvms.go
jeff-roche: 1. I'd like to see this moved to the SNO Folder since these are all SNO specific tests
2. I'd like to see this file broken apart into multiple files based on logical grouping

Maybe these groupings?
- PVCs
- LVs
- LVM Cluster
- Config
- etc.

test/integration/qe-tests/lvms_utils.go
jeff-roche: You should create a generic "utils" folder and add this there

jeff-roche (changes requested):

PR #1937, opened 2026-01-13
lvmcluster: Clean up stale node finalizers and annotations

jlhuilier-1a:
LVMVolumeGroup objects accumulate annotations with the prefix
wiped.devices.lvms.openshift.io/ (one per node) that are never cleaned up
when nodes are deleted, causing annotations to exceed Kubernetes' 256KB
limit and resulting in vg-manager reconciliation errors.

Add cleanupNodeAnnotations() to remove stale node annotations and move
existing finalizer cleanup from poll-based LVMCluster reconciliation to the
dedicated node removal controller. This provides event-driven cleanup at the
precise moment of LVMVolumeGroupNodeStatus deletion, eliminating race
conditions and improving efficiency by targeting the specific deleted node
rather than scanning all VolumeGroups on every reconciliation.

### DIFF
--- a/internal/controllers/lvmcluster/controller.go
+++ b/internal/controllers/lvmcluster/controller.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


qJkee:
closing because bugfix already merged making this PR obsolete

PR #1965, opened 2026-01-21
OCPEDGE-2326: add storageclass configuration options to lvms

Neilhamza:
this PR completes jiras:
https://issues.redhat.com/browse/OCPEDGE-2326
https://issues.redhat.com/browse/OCPEDGE-2327
https://issues.redhat.com/browse/OCPEDGE-2328

Adds configurable StorageClassOptions at the DeviceClass level in LVMCluster, allowing users to control how LVMS-generated StorageClasses are created while respecting Kubernetes immutability guarantees.

Unit tests and e2e coverage will be added in a follow-up PR.

Changes
API

Added StorageClassOptions to DeviceClass with optional:

reclaimPolicy

volumeBindingMode

additionalParameters

additionalLabels

Webhook

Validates StorageClassOptions on create/update

Warns on conflicts with LVMS-owned parameters and labels

Validates label keys/values per Kubernetes rules

Rejects updates to immutable StorageClass fields after creation

Controller

Uses CreateOrUpdate for StorageClass reconciliation

Sets immutable fields at creation time only

Supports day-2 updates for metadata-only fields (additionalLabels)

Ownership-aware label pruning ensures spec is the source of truth

No delete/recreate behavior for StorageClasses

Immutability Model

StorageClass behavior fields are immutable in Kubernetes and are configurable only at creation time.
Metadata-only fields (labels) are safely reconciled day-2 without impacting existing PVCs.

Example
kind: LVMCluster
metadata:
  name: lvmcluster
spec:
  storage:
    deviceClasses:
    - name: vg1
      storageClassOptions:
        reclaimPolicy: Retain
        volumeBindingMode: Immediate
        additionalParameters:
          provisioner-secret-name: my-secret
        additionalLabels:
          environment: production

### DIFF
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -21,7 +21,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -49,6 +49,11 @@ func (in *DeviceClass) DeepCopyInto(out *DeviceClass) {
 		in, out := &in.StorageClassOptions, &out.StorageClassOptions
 		*out = new(StorageClassOptions)
 		(*in).DeepCopyInto(*out)
+	}
+	if in.DeviceDiscoveryPolicy != nil {
+		in, out := &in.DeviceDiscoveryPolicy, &out.DeviceDiscoveryPolicy
+		*out = new(DeviceDiscoveryPolicySpec)
+		**out = **in
 	}
 }
 


api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -543,3 +566,141 @@ func (v *lvmClusterValidator) verifyMetadataSize(l *LVMCluster) ([]string, error
...
+				break
+			}
+		}
+
+		if oldDC == nil {
+			continue
+		}
+
+		scName := constants.StorageClassPrefix + newDC.Name
+		sc := &storagev1.StorageClass{}
```
qJkee: Where do you use it? I can't find anywhere

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -543,3 +566,141 @@ func (v *lvmClusterValidator) verifyMetadataSize(l *LVMCluster) ([]string, error
...
+		}
+
+		oldOpts := oldDC.StorageClassOptions
+		newOpts := newDC.StorageClassOptions
+
+		if oldOpts == nil && newOpts == nil {
+			continue
+		}
+
+		if oldOpts != nil && newOpts != nil {
```
qJkee: You already checked this at line 657

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -543,3 +566,141 @@ func (v *lvmClusterValidator) verifyMetadataSize(l *LVMCluster) ([]string, error
...
+			if oldOpts.ReclaimPolicy != nil && newOpts.ReclaimPolicy != nil {
```
qJkee: you can use `reflect.DeepEqua`l to avoid this if branching. Also, by using it, you can avoid `else` branch as well, because if old and new one is not equal = that means that there are changes that we reject

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -89,13 +115,15 @@ func (s topolvmStorageClass) EnsureDeleted(r Reconciler, ctx context.Context, lv
...
+			// Already being deleted - watch event will trigger next reconcile
+			logger.V(2).Info("StorageClass deletion in progress", "name", scName)
+			continue
 		}
 
 		if err := r.Delete(ctx, sc); err != nil {
 			return fmt.Errorf("failed to delete StorageClass %s: %w", scName, err)
 		}
-		logger.Info("initiated StorageClass deletion")
+		logger.V(2).Info("StorageClass deleted", "name", scName)
```
qJkee: this does not mean that it's deleted. It means that we started it's deletion

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -153,3 +219,59 @@ func (s topolvmStorageClass) getTopolvmStorageClasses(r Reconciler, ctx context.
 	}
 	return sc
 }
+
+func (s topolvmStorageClass) applyAdditionalLabels(
```
qJkee: Why do we need this? Why do we need to store additional labels in annotation?

api/v1alpha1/lvmcluster_webhook.go
Neilhamza: The sc variable is required for the Get() call right after, which is used to check whether the StorageClass already exists.
This existence check gates the immutability validation:
– If the StorageClass does not exist yet (NotFound), immutability checks are skipped
– If the StorageClass exists, updates to storageClassOptions are rejected

api/v1alpha1/lvmcluster_webhook.go
Neilhamza: true! updated code

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -543,3 +566,141 @@ func (v *lvmClusterValidator) verifyMetadataSize(l *LVMCluster) ([]string, error
...
+						"reclaimPolicy cannot be changed for device class %q after StorageClass creation (old: %s, new: %s)",
+						newDC.Name, *oldOpts.ReclaimPolicy, *newOpts.ReclaimPolicy)
+				}
+			} else if (oldOpts.ReclaimPolicy == nil) != (newOpts.ReclaimPolicy == nil) {
+				return fmt.Errorf(
+					"reclaimPolicy cannot be changed for device class %q after StorageClass creation",
+					newDC.Name)
+			}
+
+			if oldOpts.VolumeBindingMode != nil && newOpts.VolumeBindingMode != nil {
```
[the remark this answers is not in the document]
Neilhamza: updated code

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -53,21 +57,44 @@ func (s topolvmStorageClass) GetName() string {
...
 	logger := log.FromContext(ctx).WithValues("resourceManager", s.GetName())
 
-	// one storage class for every deviceClass based on CR is created
-	topolvmStorageClasses := s.getTopolvmStorageClasses(r, ctx, cluster)
-	for _, sc := range topolvmStorageClasses {
-		// we anticipate no edits to storage class
+	desiredStorageClasses := s.getTopolvmStorageClasses(r, ctx, cluster)
+
+	for _, desired := range desiredStorageClasses {
+		sc := &storagev1.StorageClass{
```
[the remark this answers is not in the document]
Neilhamza: true! updated code!

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -53,21 +57,44 @@ func (s topolvmStorageClass) GetName() string {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: desired.Name,
+			},
+		}
+
 		result, err := cutil.CreateOrUpdate(ctx, r, sc, func() error {
 			labels.SetManagedLabels(r.Scheme(), sc, cluster)
+			s.applyAdditionalLabels(sc, cluster, desired.Name)
+
+			if sc.CreationTimestamp.IsZero() {
```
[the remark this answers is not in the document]
Neilhamza: changed to 
if sc.ResourceVersion == "" 
for detecting creation vs update

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -543,3 +566,122 @@ func (v *lvmClusterValidator) verifyMetadataSize(l *LVMCluster) ([]string, error
...
+
+	lvmsManagedLabels := map[string]bool{
+		constants.AppKubernetesPartOfLabel:    true,
+		constants.AppKubernetesNameLabel:      true,
+		constants.AppKubernetesManagedByLabel: true,
+		constants.AppKubernetesComponentLabel: true,
+	}
+
+	for _, deviceClass := range l.Spec.Storage.DeviceClasses {
+		if deviceClass.StorageClassOptions == nil {
```
qJkee: what if they were exist and now deleted?

internal/controllers/lvmcluster/controller.go
```diff
@@ -414,6 +433,115 @@ func (r *Reconciler) checkForDeletedDeviceClasses(ctx context.Context, instance
...
+				logger.Info("failed to delete PV, continuing", "pv", pv.Name, "error", err)
+			} else {
+				deletedPVs++
+				logger.V(1).Info("requested deletion of orphaned Retain PV", "pv", pv.Name, "storageClass", pv.Spec.StorageClassName)
+			}
+		}
+	}
+
+	// Delete LogicalVolumes
+	lvList := &topolvmv1.LogicalVolumeList{}
```
qJkee: topolvm logical volumes managed by topolvm controllers and will be deleted automatically when lvmcluster is being deleted

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
Neilhamza: Right now (current implemntation) LVMS uses a partial ownership model for StorageClass labels: we only manage the keys coming from additionalLabels, and the annotation tracks which keys LVMS previously applied so we can safely prune them when removed from the spec without touching unrelated labels added by admins or external tooling.

The alternative (to implement your suggestion) would be strict ownership, where LVMS overwrites the entire labels map from the CR on every reconcile (no annotation tracking needed), but that would drop any non-LVMS labels set directly on the StorageClass. 
The first approach keeps day-2 reconciliation safe while avoiding unintended deletion of user/policy-applied labels.

can i have your suggestion on how to proceed?

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -129,6 +153,42 @@ func (s topolvmStorageClass) getTopolvmStorageClasses(r Reconciler, ctx context.
 	for _, deviceClass := range lvmCluster.Spec.Storage.DeviceClasses {
 		scName := GetStorageClassName(deviceClass.Name)
 
+		volumeBindingMode := storagev1.VolumeBindingWaitForFirstConsumer
+		deletePolicy := corev1.PersistentVolumeReclaimDelete
```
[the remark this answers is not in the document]
Neilhamza: Changed to reclaimPolicy := ptr.To(corev1.PersistentVolumeReclaimDelete) instead of intermediate variable

internal/controllers/lvmcluster/resource/topolvm_storageclass.go
```diff
@@ -51,23 +55,49 @@ func (s topolvmStorageClass) GetName() string {
 //+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;create;delete;watch;list;update;patch
 
 func (s topolvmStorageClass) EnsureCreated(r Reconciler, ctx context.Context, cluster *lvmv1alpha1.LVMCluster) error {
+	// During deletion, do not touch StorageClasses (prevents reconcile churn during finalizer cleanup)
+	if !cluster.GetDeletionTimestamp().IsZero() {
```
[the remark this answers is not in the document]
Neilhamza: Removed

api/v1alpha1/lvmcluster_webhook.go
Neilhamza: completely rewrote the immutability check in api/v1alpha1/lvmcluster_webhook.go:

   removed:
  - The StorageClass GET call (the r.Get(ctx, types.NamespacedName{Name: scName}, sc) 
  - Any dependency on checking the actual StorageClass object in the cluster

   added:
  1. immutableSCFields struct (lines 623-629) - holds only the immutable fields we care about
  2. effectiveImmutable() helper (lines 631-661) - extracts immutable fields from a DeviceClass and applies defaults (so nil -> WaitForFirstConsumer, nil -> Delete, etc.)
  3. verifyImmutableStorageClassFields() rewrite (lines 663-693) - now compares old CR spec vs new CR spec only, not cluster objects:

PR #2383, opened 2026-05-19
OCPEDGE-2612: Automate OCP-88799-Verify reclaimPolicy and volumeBindingMode combinations

mmakwana30:
Jira: https://redhat.atlassian.net/browse/OCPEDGE-2612 

Logs: https://privatebin.corp.redhat.com/?f3ebe41a5a3d0fdc#BfSzHS9X6eCKwBC8EsZ17D8Bv29RestR6KEK9UwtjSAF

## Summary by CodeRabbit

* **Tests**
  * Added integration tests validating LVMS behavior for combinations of reclaimPolicy and volumeBindingMode. Covers immediate-binding with automatic reclamation and delayed binding (WaitForFirstConsumer) with retained/manual reclamation; exercises PVC/pod lifecycles, verifies PV bind/unbind states, and confirms cleanup of retained volumes.

* **Chores**
  * Added a parameterized StorageClass template and test utilities to create StorageClasses used by the new test scenarios.

PR #2410, opened 2026-05-25
OCPEDGE-2613: Automate OCP-88798-Verify StorageClassOptions additionalParameters and Labels are applied to StorageClass

mmakwana30:
Logs: https://privatebin.corp.redhat.com/?261dfb111305351e#BvMzA2QwxGPupKVTb1o4C6GWM7aD4sN9BF3P5NCXmMAP

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive integration test validating device-class parameters and labels end-to-end: creation-time validation against reserved keys, dynamic StorageClass generation and discovery, day‑2 immutability and patch behavior for parameters, label updates and pruning, rejection of invalid/disallowed labels, and full PVC/pod mount, write/read and cleanup verification.

### DIFF
--- a/test/integration/qe_tests/lvms.go
+++ b/test/integration/qe_tests/lvms.go
@@ -5037,6 +5037,364 @@ spec:
 		o.Expect(output2).To(o.ContainSubstring("new-data-after-restore"))
 		logf("Successfully read new data from restored volume\n")
 	})
+
+	g.It("Author:mmakwana-High-88798-[OTP][LVMS] Verify StorageClassOptions additionalParameters and additionalLabels are applied to StorageClass [Disruptive]", g.Label("SNO", "MNO", "Serial"), func() {
+
+		g.By("#. Get list of available block devices/disks attached to all worker nodes")
+		freeDiskNameCountMap, err := getListOfFreeDisksFromWorkerNodes(tc)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		if len(freeDiskNameCountMap) < 1 {
+			g.Skip("Skipped: Cluster's Worker nodes does not have minimum required free block devices/disks attached")
+		}
+
+		workerNodes, err := getWorkersList()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		workerNodeCount := len(workerNodes)
+
+		var diskName string
+		isDiskFound := false
+		for disk, count := range freeDiskNameCountMap {
+			if count == int64(workerNodeCount) {
+				diskName = disk
+				isDiskFound = true
+				delete(freeDiskNameCountMap, diskName)
+				break
+			}
+		}
+		if !isDiskFound {
+			g.Skip("Skipped: All Worker nodes does not have a free block device/disk with same name attached")
+		}
+
+		g.By("#. Copy and save existing LVMCluster configuration in JSON format")
+		originLVMClusterName, err := getLVMClusterName(lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originLVMJSON, err := getLVMClusterJSON(originLVMClusterName, lvmsNamespace)
+		logf("Original LVMCluster saved: %s\n", originLVMClusterName)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("#. Delete existing LVMCluster resource")
+		deleteSpecifiedResource("lvmcluster", originLVMClusterName, lvmsNamespace)
+
+		defer func() {
+			exists, _ := resourceExists("lvmcluster", originLVMClusterName, lvmsNamespace)
+			if !exists {
+				logf("Restoring original LVMCluster from saved JSON...\n")
+				if err := createLVMClusterFromJSON(originLVMJSON); err != nil {
+					logf("Warning: Failed to restore LVMCluster from JSON: %v\n", err)
+				}
+			}
+			if err := waitForLVMClusterReady(originLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout); err != nil {
+				logf("Warning: LVMCluster did not become ready: %v\n", err)
+			}
+		}()
+
+		g.By("#. Test topolvm.io/device-class parameter rejection at creation time")
+		newLVMClusterName := "test-lvmcluster-88798"
+		deviceClassName := "vg1"
+		diskPath := "/dev/" + diskName
+
+		invalidParamClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: %s
+  namespace: %s
+spec:
+  storage:
+    deviceClasses:
+    - name: %s
+      fstype: xfs
+      thinPoolConfig:
+        name: thin-pool-1
+        sizePercent: 90
+        overprovisionRatio: 10
+      deviceSelector:
+        paths:
+        - %s
+      storageClassOptions:
+        additionalParameters:
+          topolvm.io/device-class: custom-value
+`, newLVMClusterName, lvmsNamespace, deviceClassName, diskPath)
+
+		err = createLVMClusterFromJSON(invalidParamClusterYAML)
+		o.Expect(err).To(o.HaveOccurred(), "LVMCluster with reserved topolvm.io/device-class param should be rejected")
+		o.Expect(err.Error()).To(o.ContainSubstring("managed by LVMS"), "Error should mention the key is managed by LVMS")
+		logf("topolvm.io/device-class parameter rejected at creation time as expected\n")
+
+		g.By("#. Test csi.storage.k8s.io/fstype parameter rejection at creation time")
+		invalidFstypeClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: %s
+  namespace: %s
+spec:
+  storage:
+    deviceClasses:
+    - name: %s
+      fstype: xfs
+      thinPoolConfig:
+        name: thin-pool-1
+        sizePercent: 90
+        overprovisionRatio: 10
+      deviceSelector:
+        paths:
+        - %s
+      storageClassOptions:
+        additionalParameters:
+          csi.storage.k8s.io/fstype: btrfs
+`, newLVMClusterName, lvmsNamespace, deviceClassName, diskPath)
+
+		err = createLVMClusterFromJSON(invalidFstypeClusterYAML)
+		o.Expect(err).To(o.HaveOccurred(), "LVMCluster with reserved csi.storage.k8s.io/fstype param should be rejected")
+		o.Expect(err.Error()).To(o.ContainSubstring("managed by LVMS"), "Error should mention the key is managed by LVMS")
+		logf("csi.storage.k8s.io/fstype parameter rejected at creation time as expected\n")
+
+		g.By("#. Create LVMCluster with storageClassOptions.additionalParameters and storageClassOptions.additionalLabels")
+		storageClassName := "lvms-" + deviceClassName
+
+		lvmClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: %s
+  namespace: %s
+spec:
+  storage:
+    deviceClasses:
+    - name: %s
+      fstype: xfs
+      thinPoolConfig:
+        name: thin-pool-1
+        sizePercent: 90
+        overprovisionRatio: 10
+      deviceSelector:
+        paths:
+        - %s
+      storageClassOptions:
+        additionalParameters:
+          custom.parameter/key1: value1
+          custom.parameter/key2: value2
+        additionalLabels:
+          team: storage-team
+          environment: test
+`, newLVMClusterName, lvmsNamespace, deviceClassName, diskPath)
+
+		err = createLVMClusterFromJSON(lvmClusterYAML)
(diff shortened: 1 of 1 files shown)


test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+        paths:
+        - %s
+      storageClassOptions:
+        additionalParameters:
+          topolvm.io/device-class: custom-value
+`, newLVMClusterName, lvmsNamespace, deviceClassName, diskPath)
+
+		err = createLVMClusterFromJSON(invalidParamClusterYAML)
+		o.Expect(err).To(o.HaveOccurred(), "LVMCluster with reserved topolvm.io/device-class param should be rejected")
+		o.Expect(err.Error()).To(o.ContainSubstring("topolvm.io/device-class"), "Error should mention the reserved key")
```
Neilhamza: could we assert on `"managed by LVMS"` instead? the current check works because the key name is echoed in the error, but it's matching on the input not the rejection reason

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+			return teamLabel == "updated-team" && envLabel == ""
+		}, LVMClusterReadyTimeout, 5*time.Second).Should(o.BeTrue())
+		logf("StorageClass label 'environment' pruned as expected\n")
+
+		g.By("#. Verify invalid label keys are rejected")
+		invalidLabelPatch := `[{"op":"replace","path":"/spec/storage/deviceClasses/0/storageClassOptions/additionalLabels","value":{"-starts-with-hyphen":"value1"}}]`
+		patchCmd = exec.Command("oc", "patch", "lvmcluster", newLVMClusterName, "-n", lvmsNamespace, "--type=json", "-p", invalidLabelPatch)
+		output, err = patchCmd.CombinedOutput()
+		o.Expect(err).To(o.HaveOccurred(), "invalid label key should be rejected")
+		o.Expect(string(output)).To(o.ContainSubstring("invalid"))
```
Neilhamza: `ContainSubstring("invalid")` is pretty broad, most k8s API errors have that word somewhere. `"additionalLabels key"` would be safer to make sure it's actually the LVMS webhook rejecting it

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+		if pvName != "" {
+			cleanupLogicalVolumeByName(pvName)
+			o.Eventually(func() bool {
+				cmd := exec.Command("oc", "get", "pv", pvName, "--ignore-not-found", "-o=jsonpath={.metadata.name}")
+				output, _ := cmd.CombinedOutput()
+				return strings.TrimSpace(string(output)) == ""
+			}, 2*time.Minute, 5*time.Second).Should(o.BeTrue())
+		}
+
+		g.By("#. Delete newly created LVMCluster resource")
```
Neilhamza, line 5389: the defers at ~5079 and ~5098 already handle test cluster deletion and original restore for both success and failure paths. i think this explicit cleanup block can be removed to avoid double cleanup. @pacevedom what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+
+		g.By("#. Copy and save existing LVMCluster configuration in JSON format")
+		originLVMClusterName, err := getLVMClusterName(lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originLVMJSON, err := getLVMClusterJSON(originLVMClusterName, lvmsNamespace)
+		logf("Original LVMCluster saved: %s\n", originLVMClusterName)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("#. Delete existing LVMCluster resource")
+		deleteSpecifiedResource("lvmcluster", originLVMClusterName, lvmsNamespace)
```
Neilhamza, line 5077: does `deleteSpecifiedResource` wait for the VG to be fully cleaned up on the node, or just for the CR to disappear? if it's just CR deletion there could be a race where the old VG is still on disk when the new LVMCluster tries to use the same device. @pacevedom @qjkee what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+
+		g.By("#. Test app.kubernetes.io/ reserved labels rejection")
+		appK8sPatch := `[{"op":"replace","path":"/spec/storage/deviceClasses/0/storageClassOptions/additionalLabels","value":{"app.kubernetes.io/name":"my-app"}}]`
+		patchCmd = exec.Command("oc", "patch", "lvmcluster", newLVMClusterName, "-n", lvmsNamespace, "--type=json", "-p", appK8sPatch)
+		output, err = patchCmd.CombinedOutput()
+		o.Expect(err).To(o.HaveOccurred(), "app.kubernetes.io/ labels should be rejected")
+		o.Expect(string(output)).To(o.ContainSubstring("operator-reserved"))
+		logf("app.kubernetes.io/ reserved label rejected as expected\n")
+
+		g.By("#. Delete Pod and PVC resources")
```
Neilhamza, line 5385: if `cleanupLogicalVolumeByName` is needed to unblock LVMCluster deletion, this should probably be in the defer instead. a mid-test failure would skip this block and the restore defer could hang. @pacevedom what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+			logf("Cleaning up test LVMCluster %s...\n", newLVMClusterName)
+			deleteLVMClusterSafely(newLVMClusterName, lvmsNamespace, deviceClassName)
+		}()
+
+		g.By("#. Wait for LVMCluster to become Ready")
+		err = waitForLVMClusterReady(newLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout)
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By("#. Get StorageClass and verify custom parameters exist")
+		cmd := exec.Command("oc", "get", "storageclass", storageClassName, "-o=jsonpath={.parameters}")
```
Neilhamza, line 5195: `TestClient` already wraps a typed controller-runtime client. using `HaveKeyWithValue("custom.parameter/key1", "value1")` on the parsed SC object would be more precise than `ContainSubstring` on raw jsonpath, which could match partial names. @pacevedom do you have a preference?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+			mountPath: mountPath,
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer func() {
+			deleteSpecifiedResource("pod", podName, testNamespace)
+		}()
+
+		g.By("#. Wait for PVC to be bound")
+		o.Eventually(func() string {
+			cmd := exec.Command("oc", "get", "pvc", pvcName, "-n", testNamespace, "-o=jsonpath={.status.phase}")
```
Neilhamza, line 5273: the PVC/Pod polling swallows errors with `output, _ := cmd.CombinedOutput()`, so a real failure (wrong namespace, wrong name) would just silently time out. the typed client gives better failure messages and `corev1.ClaimBound` avoids string comparison. @pacevedom do you have a preference?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+
+		err = createLVMClusterFromJSON(invalidFstypeClusterYAML)
+		o.Expect(err).To(o.HaveOccurred(), "LVMCluster with reserved csi.storage.k8s.io/fstype param should be rejected")
+		o.Expect(err.Error()).To(o.ContainSubstring("csi.storage.k8s.io/fstype"), "Error should mention the reserved key")
+		logf("csi.storage.k8s.io/fstype parameter rejected at creation time as expected\n")
+
+		g.By("#. Create LVMCluster with storageClassOptions.additionalParameters and storageClassOptions.additionalLabels")
+		storageClassName := "lvms-" + deviceClassName
+
+		lvmClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
```
Neilhamza, line 5155: since `TestClient` has the lvmv1alpha1 scheme registered, might be worth using Go structs for the valid LVMCluster creation. if an API field gets renamed the YAML template would silently create a cluster without it and the test could pass for the wrong reason. for the rejection tests YAML is fine since you're testing invalid input. @pacevedom what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,363 @@ spec:
...
+		o.Expect(scLabels).To(o.ContainSubstring("environment"))
+		o.Expect(scLabels).To(o.ContainSubstring("test"))
+		o.Expect(scLabels).To(o.ContainSubstring("owned-by.topolvm.io/name"))
+		logf("StorageClass %s has custom labels as expected\n", storageClassName)
+
+		g.By("#. Verify StorageClass can be retrieved by additionalLabels")
+		cmd = exec.Command("oc", "get", "storageclass", "-l", "team=storage-team", "-o=jsonpath={.items[*].metadata.name}")
+		output, err = cmd.CombinedOutput()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(strings.Fields(string(output))).To(o.ContainElement(storageClassName), "Expected StorageClass %s in label selector results", storageClassName)
```
Neilhamza, line 5223: `ContainElement(storageClassName)` would also pass if unrelated StorageClasses share the same label. might want to check the result count or exact match if this label is supposed to be unique to the test

Neilhamza:
thanks for the PRs @mmakwana30, well put together! the reserved key rejection tests (especially `csi.storage.k8s.io/fstype` and `owned-by.topolvm.io/` prefix) and the SSA label pruning validation are nice additions that weren't covered in the e2e suite.

a few suggestions across this PR, #2416, and #2383 since they all cover the same feature area:

**shared patterns:** the disruptive save-delete-create-restore lifecycle is copy-pasted between #2410 and #2416. would be easier to maintain as a shared helper so a cleanup bug only needs fixing once.

**typed client:** `tc.Clientset` is already used ~196 times in this file but these PRs use `exec.Command("oc")` for all verification. the typed client gives better error messages on failure, avoids string comparison (e.g. `"Bound"` vs `corev1.ClaimBound`), and doesn't spawn a subprocess every 5 seconds inside `Eventually` loops.

**magic strings:** `"Bound"`, `"Pending"`, `"Retain"`, `"Immediate"` etc. are all available as Go constants from the k8s API. using the constants catches typos at compile time.

**cleanup pattern:** some tests use defer + explicit cleanup at the end which means both run on success. picking one pattern (just defers) keeps it simpler.

**#2383 specifically:** the `Eventually` check for Pending doesn't prove the PVC *stays* Pending. a `Consistently` for a few seconds would confirm WaitForFirstConsumer is actually holding back the binding (the e2e suite does this at `storage_class_options_test.go:240`).

see inline comments on each PR for specifics.

test/integration/qe_tests/lvms.go
mmakwana30: similar response [here ](https://github.com/openshift/lvm-operator/pull/2416/changes#r3343539816)

test/integration/qe_tests/lvms.go
mmakwana30: see response [here](https://github.com/openshift/lvm-operator/pull/2383/changes#r3341759747)

test/integration/qe_tests/lvms.go
mmakwana30: Added

test/integration/qe_tests/lvms.go
mmakwana30: deleteSpecifiedResource only waits for the CR to disappear. But waitForLVMClusterReady on the new LVMCluster won't pass until the VG is fully set up, so if there is a VG conflict, the LVMCluster wouldn't become Ready and the test would fail.

test/integration/qe_tests/lvms.go
mmakwana30: response [here](https://github.com/openshift/lvm-operator/pull/2383/changes#r3341759747)

test/integration/qe_tests/lvms.go
mmakwana30: Added

test/integration/qe_tests/lvms.go
mmakwana30: Added in defer block

test/integration/qe_tests/lvms.go
mmakwana30: The test is checking that our StorageClass can be found using the label selector, not that the label is unique. ContainElement(storageClassName) confirms exactly that.. the label query returns our expected StorageClass in the results. What do you say?

test/integration/qe_tests/lvms.go
mmakwana30: Added error logging for both PVC and Pod polling loops.

pacevedom:
/retitle NO-JIRA: Automate OCP-88798-Verify StorageClassOptions additionalParameters and Labels are applied to StorageClass

/override cy/prow/snyk-deps

pacevedom:
/retitle [OCPEDGE-2613](https://redhat.atlassian.net/browse/OCPEDGE-2613): Automate OCP-88798-Verify StorageClassOptions additionalParameters and Labels are applied to StorageClass

pacevedom (approved):

PR #2416, opened 2026-05-25
OCPEDGE-2612: Automate OCP-88797 Verify StorageClassOptions reclaimPolicy and volumeBindingMode are applied to StorageClass

mmakwana30:
Logs: https://privatebin.corp.redhat.com/?ca483961ee84465d#8gSCWxw4SAJHChcVfMrbBUNLxnf3PsW6FRqFZw6WUPmv

## Summary by CodeRabbit

* **Tests**
  * Added a disruptive LVMS integration test that enforces storage class options (reclaimPolicy=Retain, immediate binding), verifies storage class fields, provisions a PVC and confirms it becomes Bound, checks the bound PV’s reclaim policy and lifecycle through Released to removal, validates immutability enforcement for storage class options, and restores the original cluster and default storage class afterward.

### DIFF
--- a/test/integration/qe_tests/lvms.go
+++ b/test/integration/qe_tests/lvms.go
@@ -5038,6 +5038,187 @@ spec:
 		logf("Successfully read new data from restored volume\n")
 	})
 
+	g.It("Author:mmakwana-High-88797-[OTP][LVMS] Verify StorageClassOptions reclaimPolicy and volumeBindingMode are applied to StorageClass [Disruptive]", g.Label("SNO", "MNO", "Serial"), func() {
+
+		g.By("#1. Copy and save existing LVMCluster configuration in JSON format")
+		originLVMClusterName, err := getLVMClusterName(lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originLVMJSON, err := getLVMClusterJSON(originLVMClusterName, lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("Original LVMCluster saved: %s\n", originLVMClusterName)
+
+		g.By("#2. Delete existing LVMCluster resource")
+		deleteSpecifiedResource("lvmcluster", originLVMClusterName, lvmsNamespace)
+
+		newLVMClusterName := "lvmcluster-88797"
+		defer func() {
+			// Delete the test LVMCluster if it exists
+			deleteLVMClusterSafely(newLVMClusterName, lvmsNamespace, volumeGroup)
+
+			exists, _ := resourceExists("lvmcluster", originLVMClusterName, lvmsNamespace)
+			if !exists {
+				logf("Restoring original LVMCluster from saved JSON...\n")
+				// Delete the StorageClass so operator can recreate with original settings
+				defaultSCName := "lvms-" + volumeGroup
+				deleteSpecifiedResource("storageclass", defaultSCName, "")
+
+				if err := createLVMClusterFromJSON(originLVMJSON); err != nil {
+					logf("Warning: Failed to restore LVMCluster from JSON: %v\n", err)
+				}
+			}
+			if err := waitForLVMClusterReady(originLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout); err != nil {
+				logf("Warning: LVMCluster did not become ready: %v\n", err)
+			}
+		}()
+
+		// Wait for operator to settle and StorageClass to be cleaned up
+		defaultSCName := "lvms-" + volumeGroup
+		o.Eventually(func() bool {
+			exists, _ := resourceExists("storageclass", defaultSCName, "")
+			return !exists
+		}, 2*time.Minute, 5*time.Second).Should(o.BeTrue(), "StorageClass should be deleted by operator after LVMCluster deletion")
+
+		g.By("#3. Create new LVMCluster with storageClassOptions (reclaimPolicy: Retain, volumeBindingMode: Immediate)")
+		newLVMClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: %s
+  namespace: %s
+spec:
+  storage:
+    deviceClasses:
+    - name: %s
+      default: true
+      fstype: xfs
+      thinPoolConfig:
+        name: thin-pool-1
+        sizePercent: 90
+        overprovisionRatio: 10
+      storageClassOptions:
+        reclaimPolicy: Retain
+        volumeBindingMode: Immediate
+`, newLVMClusterName, lvmsNamespace, volumeGroup)
+
+		err = createLVMClusterFromJSON(newLVMClusterYAML)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("Created new LVMCluster with storageClassOptions\n")
+
+		g.By("#4. Wait for LVMCluster to become Ready")
+		err = waitForLVMClusterReady(newLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("LVMCluster %s is Ready\n", newLVMClusterName)
+
+		g.By("#5. Get the created StorageClass and verify reclaimPolicy is Retain")
+		cmd := exec.Command("oc", "get", "storageclass", defaultSCName, "-o=jsonpath={.reclaimPolicy}")
+		output, err := cmd.CombinedOutput()
+		o.Expect(err).NotTo(o.HaveOccurred())
+		o.Expect(strings.TrimSpace(string(output))).To(o.Equal("Retain"))
+		logf("StorageClass %s has reclaimPolicy: Retain as expected\n", defaultSCName)
+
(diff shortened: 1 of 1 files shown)


test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+
+		g.By("#1. Copy and save existing LVMCluster configuration in JSON format")
+		originLVMClusterName, err := getLVMClusterName(lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		originLVMJSON, err := getLVMClusterJSON(originLVMClusterName, lvmsNamespace)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("Original LVMCluster saved: %s\n", originLVMClusterName)
+
+		g.By("#2. Delete existing LVMCluster resource")
+		deleteLVMClusterSafely(originLVMClusterName, lvmsNamespace, volumeGroup)
```
Neilhamza: `deleteLVMClusterSafely` strips all finalizers before deleting, which skips the entire operator cleanup chain. normally the LVMCluster finalizer triggers `processDelete` in the lvmcluster controller, which deletes the LVMVolumeGroups, which in turn triggers vgmanager `processDelete` on each node to remove the thin pool, delete the VG (`vgremove`), clean up lvmd config, and remove the node status. stripping finalizers means none of that runs — the VG and thin pool stay on disk, lvmd config is stale, and the next LVMCluster on the same disk will likely fail. should use `deleteSpecifiedResource` instead and let the operator do its cleanup, or at minimum wait for the VG to be gone from the node after deletion. same concern as https://github.com/openshift/lvm-operator/pull/2410#discussion_r3329989271 @pacevedom @qjkee what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+				}
+			}
+			if err := waitForLVMClusterReady(originLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout); err != nil {
+				logf("Warning: LVMCluster did not become ready: %v\n", err)
+			}
+		}()
+
+		// Delete the existing StorageClass so new one can be created with different settings
+		defaultSCName := "lvms-" + volumeGroup
+		deleteSpecifiedResource("storageclass", defaultSCName, "")
```
Neilhamza: the SC is deleted here but if the old operator reconcile loop hasn't fully stopped, it could recreate the SC with old settings before the new cluster takes over. might be worth waiting for the operator to settle first. @pacevedom what do you think?

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+			output, err := cmd.CombinedOutput()
+			if err != nil {
+				logf("oc error checking PV deletion: %v: %s\n", err, string(output))
+				return false
+			}
+			return strings.TrimSpace(string(output)) == ""
+		}, 2*time.Minute, 5*time.Second).Should(o.BeTrue())
+		logf("Orphaned PV %s deleted successfully\n", pvName)
+
+		g.By("#15. Delete newly created LVMCluster resource")
```
Neilhamza, line 5196: the defer at ~5056 already handles test cluster deletion and original restore. this explicit cleanup (steps 15-17) duplicates that, so on success both run. same concern as https://github.com/openshift/lvm-operator/pull/2410#discussion_r3329989270

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("Created new LVMCluster with storageClassOptions\n")
+
+		g.By("#4. Wait for LVMCluster to become Ready")
+		err = waitForLVMClusterReady(newLVMClusterName, lvmsNamespace, LVMClusterReadyTimeout)
+		o.Expect(err).NotTo(o.HaveOccurred())
+		logf("LVMCluster %s is Ready\n", newLVMClusterName)
+
+		g.By("#5. Get the created StorageClass and verify reclaimPolicy is Retain")
+		cmd := exec.Command("oc", "get", "storageclass", defaultSCName, "-o=jsonpath={.reclaimPolicy}")
```
Neilhamza, line 5112: this file already uses `tc.Clientset` for PVC, PV, and SC gets in other tests (e.g. `tc.Clientset.StorageV1().StorageClasses().Get()` at line 349, `tc.Clientset.CoreV1().PersistentVolumeClaims().Get()` throughout). the `exec.Command("oc", "get", ... "-o=jsonpath")` pattern here does the same thing but with string comparison instead of typed fields, and swallows errors in the Eventually loops. for example this block could be:

```go
sc, err := tc.Clientset.StorageV1().StorageClasses().Get(context.TODO(), defaultSCName, metav1.GetOptions{})
o.Expect(err).NotTo(o.HaveOccurred())
o.Expect(string(*sc.ReclaimPolicy)).To(o.Equal("Retain"))
o.Expect(string(*sc.VolumeBindingMode)).To(o.Equal("Immediate"))
```

same applies to the PVC phase, PV reclaimPolicy, and PV phase checks below. same concern as https://github.com/openshift/lvm-operator/pull/2410#discussion_r3329989275

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+		g.By("#7. Create a PVC using the StorageClass (without creating a pod)")
+		pvcName := "test-pvc-88797"
+		err = createPVCWithOC(pvcConfig{
+			name:             pvcName,
+			namespace:        testNamespace,
+			storageClassName: defaultSCName,
+			storage:          "2Gi",
+		})
+		o.Expect(err).NotTo(o.HaveOccurred())
+		defer deleteSpecifiedResource("pvc", pvcName, testNamespace)
```
Neilhamza, line 5134: this defer deletes the PVC but step 10 at line 5155 also deletes it explicitly. harmless because of `--ignore-not-found` but adds noise

test/integration/qe_tests/lvms.go
```diff
@@ -5037,6 +5037,184 @@ spec:
...
+				logf("Warning: LVMCluster did not become ready: %v\n", err)
+			}
+		}()
+
+		// Delete the existing StorageClass so new one can be created with different settings
+		defaultSCName := "lvms-" + volumeGroup
+		deleteSpecifiedResource("storageclass", defaultSCName, "")
+
+		g.By("#3. Create new LVMCluster with storageClassOptions (reclaimPolicy: Retain, volumeBindingMode: Immediate)")
+		newLVMClusterYAML := fmt.Sprintf(`apiVersion: lvm.topolvm.io/v1alpha1
```
Neilhamza, line 5082: same as https://github.com/openshift/lvm-operator/pull/2410#discussion_r3329989277 — using a Go struct instead of YAML for the valid LVMCluster creation would catch API field renames at compile time

Neilhamza:
thanks for this @mmakwana30! i left a consolidated review comment on #2410 that covers suggestions across all three PRs: https://github.com/openshift/lvm-operator/pull/2410#issuecomment-4586533096

see the inline comments here for anything specific to this PR.

test/integration/qe_tests/lvms.go
mmakwana30: Added deleteSpecifiedResource instead of deleteLVMClusterSafely

test/integration/qe_tests/lvms.go
mmakwana30: Added a wait for the StorageClass to be fully deleted before creating the new LVMCluster. This will ensure  operator has settled and won't recreate the SC with old settings

test/integration/qe_tests/lvms.go
mmakwana30: using oc client as discussed.

test/integration/qe_tests/lvms.go
mmakwana30: using oc client as discussed.

test/integration/qe_tests/lvms.go
mmakwana30: response [here ](https://github.com/openshift/lvm-operator/pull/2383/changes#r3341717278)

test/integration/qe_tests/lvms.go
mmakwana30: If the test fails before step 15, the defer handles cleanup and restores the original cluster. If the test passes, steps 15-17 run and step 17 verifies the restored settings. The defer still runs but the exists check skips double-restore. This follows the same pattern used across the test suite (e.g., 71012, 66241).

pacevedom:
/retitle NO-JIRA: Automate OCPBUGS-88797 Verify StorageClassOptions reclaimPolicy and volumeBindingMode are applied to StorageClass

pacevedom:
/override ci/prow/snyk-deps

Unrelated error

pacevedom:
/retitle [OCPEDGE-2612](https://redhat.atlassian.net/browse/OCPEDGE-2612): Automate OCP-88797 Verify StorageClassOptions reclaimPolicy and volumeBindingMode are applied to StorageClass

pacevedom (approved):

PR #2434, opened 2026-05-27
NO-JIRA: Update owners aliases

pacevedom:


OWNERS_ALIASES
```diff
@@ -7,6 +7,8 @@ aliases:
   - brandisher
   - qJkee
   - jaypoulz
+  - pacevedom
   edge-enablement-qe:
```
brandisher: The old alias names need updating.

pacevedom: Done

PR #2438, opened 2026-05-27
[release-4.19] NO-JIRA: Limit go test parallelism to prevent OOM in CI

pacevedom:
The Konflux integration test pipeline runs with a memory limit from the tenant LimitRange. With default parallelism, multiple packages compile heavy dependencies simultaneously, exceeding the limit and triggering OOM kills. Limit to 2 parallel packages.


pacevedom:
/override ci/prow/snyk-deps

Unrelated error to this PR.

PR #2494, opened 2026-06-03
OCPBUGS-86959: Fix storageClassOptions XValidation immutability bypass

Neilhamza:
## Summary
- Fix XValidation immutability bypass when `storageClassOptions` transitions from nil to non-nil on an existing LVMCluster
- Add `+kubebuilder:default={}` on the `StorageClassOptions` field so defaults are always populated on creation
- Add webhook validation for the upgrade path (existing CRs without `storageClassOptions` populated)

## Root Cause
Kubernetes skips CEL transition rules (`oldSelf == self`) when `oldSelf` doesn't exist at the field path. When `storageClassOptions` was nil and the user added it for the first time with non-default immutable values (e.g. `reclaimPolicy: Retain`), the XValidation was silently skipped. The operator then couldn't reconcile the StorageClass because Kubernetes treats SC fields as immutable.

## Fix
1. **CRD default**: `+kubebuilder:default={}` on `StorageClassOptions` ensures the API server always populates it with defaults on CREATE. Then `oldSelf` is always present on subsequent UPDATEs and the existing XValidation rules work correctly.
2. **Webhook guard**: `validateStorageClassOptionsUpgrade()` rejects non-default immutable values when `storageClassOptions` transitions from nil to non-nil. This covers existing CRs created before the default was added (upgrade path).

## Affected Fields
All three immutable fields: `reclaimPolicy`, `volumeBindingMode`, `additionalParameters`.

## Test plan
- [x] Unit tests: 9 new tests covering nil-to-non-nil transition (3 rejection, 4 acceptance, 1 default verification, 1 existing test assertion update)
- [x] All existing unit tests pass (70 API tests, 0 failures)
- [x] Verified on live single-node OpenShift cluster:
  - Create without storageClassOptions → defaults populated automatically
  - Try to add reclaimPolicy: Retain → rejected by XValidation
  - Try to add volumeBindingMode: Immediate → rejected by XValidation
  - Try to add additionalParameters → rejected by XValidation
  - Add additionalLabels → accepted (mutable field)
  - Operator healthy, no reconciliation errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added coverage for StorageClassOptions defaulting on create and comprehensive upgrade scenarios (nil→non-nil), including accepted and rejected transitions and behavior when adding new device classes.

* **Chores**
  * StorageClassOptions now defaults to an empty object when omitted.
  * Validation on updates enforces immutability of specific storage-class settings while allowing permitted additions.


pacevedom:
Hey @Neilhamza, we might need to rebase the PR again and also run [make bundle](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_lvm-operator/2494/pull-ci-openshift-lvm-operator-main-precommit-check/2062147364784705536):
```
There are unexpected changes to the tree when running 'make bundle'. Please
run these commands locally and double-check the Git repository for unexpected changes which may
need to be committed.
```

api/v1alpha1/lvmcluster_webhook.go
```diff
@@ -200,6 +201,10 @@ func (v *lvmClusterValidator) ValidateUpdate(_ context.Context, oldLVMCluster, l
 		return warnings, err
 	}
 
+	if err := validateStorageClassOptionsUpgrade(oldLVMCluster.Spec.Storage.DeviceClasses, l.Spec.Storage.DeviceClasses); err != nil {
+		return warnings, err
```
qJkee: can we wrap error with fmt.Errorf to provide extra context? We are doing that everywhere

qJkee (changes requested):
Just one minor thing, otherwise LGTM

api/v1alpha1/lvmcluster_webhook.go
Neilhamza: done

PR #2549, opened 2026-06-05
OCPEDGE-2522: feat(RAID): add basic RAID support

qJkee:
Introduce RAIDConfig in the LVMVolumeGroup API to allow configuring LVM RAID levels (raid1, raid4, raid5, raid6, raid10). The vg-manager validates device counts and passes RAID-specific lvcreate options through the LVMD configuration.

## Summary by CodeRabbit

* **New Features**
  * Added native LVM RAID support for volume groups (raid1, raid4, raid5, raid6, raid10) with mirrors, stripes, and stripe-size options; RAID is immutable after creation and mutually exclusive with thin pool configuration.
  * Device discovery mode is adjusted automatically when RAID is configured.
  * Volume groups with insufficient devices for the requested RAID now transition to Failed and emit a warning.

* **Tests**
  * Added unit and integration tests covering RAID option conversion, device-count validation, config generation, lifecycle reconciliation, and cleanup.


qJkee:
@CodeRabbit review

internal/controllers/vgmanager/controller.go
```diff
@@ -216,6 +216,9 @@ func (r *Reconciler) reconcile(
 	}
 
 	effectivePolicy := ptr.Deref(volumeGroup.Spec.DeviceDiscoveryPolicy, lvmv1alpha1.DeviceDiscoveryPolicyDynamic)
+	if volumeGroup.Spec.RAIDConfig != nil {
+		effectivePolicy = lvmv1alpha1.DeviceDiscoveryPolicyStatic
```
Neilhamza, line 220: Nit: `status.go` reads `vg.Spec.DeviceDiscoveryPolicy` directly, so when RAID overrides the effective policy to Static here, the reported status could say RuntimeDynamic instead of RuntimeStatic.

internal/controllers/vgmanager/raid.go
```diff
@@ -0,0 +1,55 @@
...
+		if rc.StripeSize != nil {
+			opts = append(opts, "--stripesize", fmt.Sprintf("%dk", rc.StripeSize.Value()/1024))
+		}
+	}
+
+	return opts
+}
+
+func validateRAIDDeviceCount(rc *lvmv1alpha1.RAIDConfig, deviceCount int) error {
+	mirrors := rc.EffectiveMirrors()
```
Neilhamza, line 37: I see the webhook already checks device counts, could this runtime validation be redundant?

api/v1alpha1/lvmvolumegroup_types.go
```diff
@@ -37,6 +38,12 @@ type LVMVolumeGroupSpec struct {
 	// +optional
 	ThinPoolConfig *ThinPoolConfig `json:"thinPoolConfig,omitempty"`
 
+	// RAIDConfig configures native LVM RAID for this volume group.
+	// Mutually exclusive with ThinPoolConfig. All fields are immutable after creation.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="oldSelf == self",message="raidConfig is immutable after creation"
+	RAIDConfig *RAIDConfig `json:"raidConfig,omitempty"`
```
Neilhamza, line 45: Field-level constraints like mirrors-only-for-raid1/raid10 are only validated by the LVMCluster webhook, not at this CRD level.

internal/controllers/vgmanager/controller.go
```diff
@@ -308,6 +311,23 @@ func (r *Reconciler) reconcile(
 
 	logger.Info("new available devices discovered", "available", devices.Available)
 
+	if volumeGroup.Spec.RAIDConfig != nil {
+		totalDevices := len(devices.Available)
```
Neilhamza, line 315: Should we also validate the actual PV count after VG creation, in case a device fails to initialize during `addDevicesToVG` and we end up with fewer PVs than the RAID level requires?

internal/controllers/vgmanager/controller.go
```diff
@@ -216,6 +216,9 @@ func (r *Reconciler) reconcile(
 	}
 
 	effectivePolicy := ptr.Deref(volumeGroup.Spec.DeviceDiscoveryPolicy, lvmv1alpha1.DeviceDiscoveryPolicyDynamic)
+	if volumeGroup.Spec.RAIDConfig != nil {
```
Neilhamza, line 219: Since the LVMCluster webhook already requires explicit device paths when RAID is configured, and `hasExplicitDevicePaths()` skips the Static logic entirely, this override would only matter for manually created LVMVolumeGroups without paths, is that intentional or could we remove it?

internal/controllers/vgmanager/raid.go
```diff
@@ -0,0 +1,55 @@
...
+
+	lvmv1alpha1 "github.com/openshift/lvm-operator/v4/api/v1alpha1"
+)
+
+func buildRAIDLVCreateOptions(rc *lvmv1alpha1.RAIDConfig) []string {
+	opts := []string{"--type", string(rc.Type)}
+
+	switch rc.Type {
+	case lvmv1alpha1.RAIDTypeRAID1:
+		opts = append(opts, "-m", fmt.Sprintf("%d", rc.EffectiveMirrors()))
```
Neilhamza: Minor nit: should this be replaced with `strconv.Itoa()`?

internal/controllers/vgmanager/controller_test.go
```diff
@@ -75,6 +75,9 @@ var _ = Describe("vgmanager controller", func() {
 				Spec: lvmv1alpha1.LVMVolumeGroupSpec{},
 			})
 		})
+		It("RAID volume group", func(ctx SpecContext) {
```
Neilhamza, line 78: Would it be worth adding a case for the error path too, like raid1 with only 1 device, to exercise the `validateRAIDDeviceCount` failure handling?

api/v1alpha1/lvmvolumegroup_types.go
Neilhamza: which is probably fine since the operator is the sole creator, but maybe worth a comment on the field noting that

internal/controllers/vgmanager/raid.go
qJkee: Because webhook can be deleted, it's nice to have some extra validation to guard users from breaking RAID clusters

api/v1alpha1/lvmvolumegroup_types.go
qJkee: Yes, it's validated via webhook and operator itself because writing such complex CEL rule will be hard to read and manage in the future

internal/controllers/vgmanager/controller.go
qJkee: If device isn't added to the LVM VG, LVM command will return error. Cluster will degrade and we indicate error. Same as non RAID clusters

internal/controllers/vgmanager/controller.go
qJkee: hasExplicitDevicePaths is checked first in status.go:84, and RAID requires explicit paths (webhook.go:686-689), so RAID VGs always hit the DeviceDiscoveryPolicyPreconfigured branch resulting to correct status type

internal/controllers/vgmanager/raid.go
qJkee: fixed

internal/controllers/vgmanager/controller_test.go
qJkee: fixed

internal/controllers/vgmanager/controller.go
qJkee: The LVMCluster webhook validates the primary path, but LVMVolumeGroup is its own CRD — a manually created one could bypass the webhook

PR #2750, opened 2026-06-24
OCPEDGE-791: Skip RapiDAST test on disconnected clusters

mmakwana30:


pacevedom (approved):
/lgtm
/hold

test/integration/qe_tests/rapidast.go
```diff
@@ -67,6 +67,39 @@ func isARMCluster() bool {
 	return strings.Contains(archs, "arm64") || strings.Contains(archs, "aarch64")
 }
 
+func isDisconnectedCluster() bool {
```
kasturinarra, line 70: can we add this in the utils so that this can be used across ?

PR #2798, opened 2026-07-01
NO-JIRA: Add script to check Go toolchain version in LVMS images

pacevedom:
During multi-image updates (e.g. Go toolchain bumps), Konflux produces intermediate catalogs where only some images carry the new toolchain. This script inspects a catalog or bundle image, extracts the LVMS-owned related images, and reports the Go version, image version tags, and source commit SHA for each binary — making it easy to verify consistency before cutting a z-stream release.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line check to inspect the Go toolchain version used in LVMS container images.
  * Supports inspecting a single image or resolving images from bundle and catalog references, including aggregating multiple bundle versions when requested.
  * Reports discovered Go binaries (or clearly indicates when none are found).
  * Outputs results as a readable table or as JSON.

### DIFF
--- a/hack/check-go-version.sh
+++ b/hack/check-go-version.sh
@@ -3,6 +3,7 @@ set -euo pipefail
 
 WORK_DIR=""
 JSON_OUTPUT=false
+ALL_VERSIONS=false
 MODE=""
 IMAGE_REF=""
 
@@ -22,6 +23,7 @@ usage() {
     echo "  --image <ref>    Check Go version directly on a single container image"
     echo "  --bundle <ref>   Inspect a bundle image and check Go version on its related images"
     echo "  --catalog <ref>  Inspect a catalog image, extract the latest bundle, and check Go versions"
+    echo "  --all            With --catalog, check the latest bundle of each minor version"
     echo "  --json           Output results as JSON"
     echo "  --help           Show this help message"
     exit 1
@@ -170,7 +172,7 @@ get_related_images_from_bundle() {
     rm -rf "${manifests}"
 }
 
-get_related_images_from_catalog() {
+extract_catalog_json() {
     local image="$1"
     local pulled_ref
     pulled_ref=$(pull_image "${image}") || return 1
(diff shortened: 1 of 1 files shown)


suleymanakbas91 (commented):
A few issues spotted in the review — none are showstoppers for a `hack/` utility, but #1 and #2 will bite real users.

**1. `set -u` + missing argument = cryptic error (lines 243-257)**

`--image`, `--bundle`, `--catalog` all reference `$2` without checking it exists. With `set -euo pipefail`, running `./check-go-version.sh --image` (no value) gives "unbound variable" instead of the usage message. Add a guard:

```bash
--image)
    [[ $# -ge 2 ]] || usage
    MODE="image"
    IMAGE_REF="$2"
    shift 2
    ;;
```

**2. Version sort breaks on non-numeric suffixes (lines 201, 219)**

`map(tonumber)` fails if a bundle version contains a pre-release suffix (e.g., `4.17.0-rc1`). The `0-rc1` segment isn't a number. Consider splitting on `-` first or stripping the suffix:

```jq
split(".") | map(split("-") | .[0] | tonumber)
```

**3. `podman create` error handling inconsistent (line 172 vs 131)**

`get_related_images_from_bundle` does bare `cid=$(podman create ...)` with no error handling, while `check_go_version_in_image` properly handles it with `|| { echo >&2 "..."; return 1; }`. Should be consistent.

**4. Container leak on early exit**

The `cleanup` trap only removes `WORK_DIR`. If the script dies mid-way (via `set -e`), containers created with `podman create` are orphaned on disk. Consider tracking container IDs and cleaning them in the trap, or using `podman create --rm`.

**5. `--all` silently ignored for non-catalog modes**

`--all` only has effect with `--catalog`, but using `--all --bundle ...` silently does nothing. Worth either warning or erroring:

```bash
if [[ "${ALL_VERSIONS}" == "true" && "${MODE}" != "catalog" ]]; then
    echo >&2 "Warning: --all only applies to --catalog mode"
fi
```

**6. Empty lines in image list (lines 278-298)**

If `get_related_images_from_bundle`/`get_related_images_from_catalog` output trailing newlines, empty strings end up in the `images` array. A guard like `[[ -z "${img}" ]] && continue` in the read loop would be safer.

pacevedom:
Addressed all comments.

pacevedom:
/override ci/prow/e2e-aws-hypershift

Unrelated failure

PR #2802, opened 2026-07-02
NO-JIRA: fix contextification doc accuracy issues

Neilhamza:
## Summary

Fixes 7 accuracy issues found by an external line-by-line audit of all contextification docs against current source code.

### Fixes

1. **architecture.md — lvmd ConfigMap claim**: Said "configured via a lvmd.yaml ConfigMap." Code uses file-based config at `/etc/topolvm/lvmd.yaml` (ConfigMap approach was reverted in PR #566, correctly documented in ADR-0009, but architecture.md still had the stale claim).

2. **"VG Manager reconciles every 30 seconds" — stale in 3 files**: AGENTS.md, architecture.md, reconciliation.md. VG Manager requeue is now conditional: RAID/Dynamic = 30s periodic; Static + explicit paths = no periodic requeue (`determineFinishedRequeue` in controller.go).

3. **architecture.md — NodeStatus cardinality**: Said "one per node per device class." Actually one per node (named after node), containing a list of per-VG status entries.

4. **known-limitations.md — bind-mount filter**: Described a `/proc/1/mountinfo` filter mechanism that does not exist in the filter chain (7 filters, none for bind mounts). Updated to note this is a user responsibility, not automated.

5. **AGENTS.md — missing controller**: Key Directories omitted `persistent-volume-claim` controller.

6. **[ADR-0010](https://redhat.atlassian.net/browse/ADR-0010) — additionalParameters mutability**: Decision drivers read as if additionalParameters are day-2 mutable. They are immutable after creation (CEL `oldSelf == self`). Only AdditionalLabels is mutable.

### Not fixed in this PR (needs team discussion)

**DeviceDiscoveryPolicy nil-default contradiction**: Controller defaults nil to Dynamic (`ptr.Deref(policy, Dynamic)`), but API godoc and webhook warning say "new VGs default to Static." The code, its own documentation, and the webhook text disagree. This is a code-internal contradiction that needs a Jira ticket and team decision, not a doc fix.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Documentation**
  * Clarified reconciliation timing so users can see when updates happen automatically versus conditionally.
  * Updated storage class guidance to note which settings can be changed without recreating resources and which are fixed after creation.
  * Refined known limitations for device selection, including that bind-mounted devices must be avoided.
  * Improved architecture and safety notes to better describe status handling, retry behavior, and controller coverage.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,7 +61,7 @@ Paths []string `json:"paths,omitempty"`
 This operator manages physical storage and performs destructive LVM operations. See [docs/core-beliefs.md](docs/core-beliefs.md) for non-negotiable invariants and [docs/conventions/](docs/conventions/) for implementation patterns.
 
 - **Data loss risk**: LVM operations can wipe disks. Verify device selectors carefully in tests.
-- **Idempotency**: controllers must handle partial states and retries safely. VG Manager reconciles every 30 seconds.
+- **Idempotency**: controllers must handle partial states and retries safely. VG Manager requeue interval depends on configuration (RAID/Dynamic: 30s periodic; Static with explicit paths: no periodic requeue).
 - **Finalizers**: LVMS uses a three-level finalizer hierarchy to prevent orphaned storage. Never skip finalizer logic. See [docs/architecture.md](docs/architecture.md) for details.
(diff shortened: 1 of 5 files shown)


pacevedom (approved):
/lgtm

PR #2862, opened 2026-07-13
NO-JIRA: feat(status): fix raid status display

qJkee:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RAID health classification across volume health states.
  * RAID arrays with partially healthy volumes are now reported as **Degraded** instead of **Failed**.
  * RAID status is now **Failed** only when all volumes are in a failed (non-partial) state.
  * Mixed combinations of failed and partial volumes now correctly show **Degraded** overall status.

### DIFF
--- a/internal/controllers/vgmanager/raid_test.go
+++ b/internal/controllers/vgmanager/raid_test.go
@@ -359,13 +359,13 @@ func TestBuildRAIDStatus(t *testing.T) {
 			expectedMinSyncPercent: ptr.To(42),
 		},
 		{
-			name: "single LV with partial health is failed",
+			name: "single LV with partial health is degraded",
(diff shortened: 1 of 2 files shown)


qJkee:
@coderabbitai review

internal/controllers/vgmanager/raid_test.go
pacevedom, line 1: Should we have a new test on the behavior that changed? For example, with 2 volumes, one with "refresh needed" and the other being partial the old code would have flagged it as failed, while the new code would report degraded.

qJkee: added a test case

pacevedom (approved):
/lgtm

PR #2914, opened 2026-07-23
OCPBUGS-90617: feat(tests): read new overprovision ratio

qJkee:
Read updated overprovision ratio value before running other tests

## Summary by CodeRabbit

* **Tests**
  * Strengthened validation of storage capacity updates after reducing LVM overprovisioning.
  * Added a convergence wait to confirm CSI-reported storage capacity decreases before running the expected PVC provisioning failure assertions.


pacevedom (approved):
/lgtm

